PR #2, opened 2014-07-04
Adding account vat on payment 7

eLBati:
Moved from https://code.launchpad.net/~domsense/account-payment/adding_account_vat_on_payment_7/+merge/196914


account_vat_on_payment/account.py
```diff
@@ -0,0 +1,433 @@
...
+                    ):
+                        vat_on_p += 1
+            if vat_on_p and vat_on_p != valid_lines:
+                raise osv.except_osv(
+                    _('Error'),
+                    _("""Can't handle VAT on payment if not every invoice
+                    is on a VAT on payment treatment"""))
+        return vat_on_p
+
+    def action_move_line_create(self, cr, uid, ids, context=None):
```
rvalyi: Sorry, cannot approve such a large method without test at the very least. If it's copied from some super method. Then I would like to see explicitly where it differs and suggest the original method is monkey patched (with new hooks made totally explicit) or patches in v8 so that it can be extended in a more modular way.

account_vat_on_payment/account.py
```diff
@@ -0,0 +1,433 @@
...
+        return res
+
+
+class account_invoice(orm.Model):
+
+    def _get_vat_on_payment(self, cr, uid, context=None):
+        return self.pool.get('res.users').browse(
+            cr, uid, uid, context).company_id.vat_on_payment
+
+    def finalize_invoice_move_lines(self, cr, uid, invoice_browse, move_lines):
```
rvalyi: same thing: quite a large method and no test...

rvalyi:
:-1: too large methods, no test -> hassle to maintain in this form

account_vat_on_payment/account.py
eLBati: It's not copied from a super method but it certainly needs tests

eLBati:
Hello @rvalyi I added the tests described at https://docs.google.com/spreadsheet/ccc?key=0Aodwq17jxF4edDJaZ2dOQkVEN0hodEtfRmpVdlg2Vnc#gid=0

.travis.yml
```diff
@@ -12,6 +12,7 @@ virtualenv:
 
 install:
   - pip install unidecode
+  - pip install egenix-mx-base
```
eLBati, line 15: see https://github.com/eLBati/account-payment/commit/047aa238304694b9359ff25bcc273307e24bcfe9

account_vat_on_payment/account.py
```diff
@@ -0,0 +1,433 @@
+# -*- coding: utf-8 -*-
```
bwrsandman: Please split this script into one script per method/class:
- account_voucher.py
- account_invoice.py
- account_move_line.py
- account_account.py
- account_tax_code.py
- account_journal.py

This script is already ridiculously long, this is exactly why we should split by model.

I'll be able to do a better review after that.

account_vat_on_payment/account.py
bwrsandman: Such a large function (with most of it in 1 to 2 levels of loops) should be split into smaller more maintainable pieces of code.

This does not look good at all.

eLBati:
Hello @bwrsandman and @rvalyi , just made one file per class and splitted `action_move_line_create` in several methods.
Thanks for your reviews

account_vat_on_payment/account_fiscal_position_view.xml
```diff
@@ -0,0 +1,48 @@
...
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
```
max3903, line 26: I don't think the header is appropriate here. This is an xml file.

rvalyi:
@eLBati it looks much better code wise (see @max3903 remark still). I've asked Anybox guys who worked on that topic if they can have a functional review. It looks like in v9 our intuitions about voucher being shit are being confirmed, that may still be acceptable in v7 still and may be we could have a better way in v8 already.

ccomb:
I'm not in favor of merging this PR into OCA, mainly because of the shadow accounts idea. I don't understand why it is necessary to create all this shadow stuffs, and I don't like the initial invoice generating no accounting entries but in the shadow accounts. An invoice should always create real accounting entries when recorded. Please tell if I did not understand fully, or maybe I don't have enough information. I also don't like the too many added fields. I'm not sure all this is necessary.
In the french accounting, when you want to record an invoice with vat on payment, you don't need an additional module, you just write the vat on different accounts (in french CoA, 4458:  waiting VAT,  instead of 4457: due VAT). Then during the reconciliation of the payment, you just transfer the correct amount from waiting VAT account to the due VAT account. IMO the real job of such a module is to compute the correct amount to transfer during the reconciliation, which may be sometimes difficult, but not more than what the accountant must do anyway.
Last year I had started a POC on this idea: https://bitbucket.org/anybox/tax_on_payment/src (with a french explanation on the wiki page). This is unfinished and it still doesn't work yet, I was just waiting a little bit of funding to finish it. But today it seems easier to find money for proprietary cms themes than for a correct accounting behaviour :-(

eLBati:
Hello @ccomb I think the 'shadow' term has been an unhappy choice, originated from the Swiss use of the module.
However, as noted at the end of the [how-to](http://planet.agilebg.com/en/2012/10/vat-on-payment-treatment-with-openerp/), the accounts to be involved are configurable.
Take for example the [Italian case](https://docs.google.com/spreadsheet/ccc?key=0Aodwq17jxF4edDJaZ2dOQkVEN0hodEtfRmpVdlg2Vnc#gid=10) (that should be similar to the French one). Here, no shadow accounts are involved, just the 'waiting VAT'.

I'd better adjust the module demo data (and some variables names) and write some test cases where 'shadow' accounts are not involved

> IMO the real job of such a module is to compute the correct amount to transfer during the reconciliation

That is exactly the purpose of this module: to compute the right amount of VAT to pay, expecially when the computation is not trivial (see [partial payments](https://docs.google.com/spreadsheet/ccc?key=0Aodwq17jxF4edDJaZ2dOQkVEN0hodEtfRmpVdlg2Vnc#gid=1) or [multi currency payments](https://docs.google.com/spreadsheet/ccc?key=0Aodwq17jxF4edDJaZ2dOQkVEN0hodEtfRmpVdlg2Vnc#gid=4))

ccomb:
The "shadow" term is indeed badly chosen, but I just tried the module and could not reproduce what I wanted to do. Please tell if I'm wrong, I wanted to do the following thing for my own use:
- Create a "VAT 20% on debits" configured on "4457 due vat" account
- Create a "VAT 20% on payment" configured on "4458 waiting vat" account (supposed to be the shadow account)
- Link the "4458 waiting vat" account to the "4457 due vat" account (through the shadow field of the account)
- Create a product of type "service", configured with the "VAT 20% on payment" tax.
- Create an invoice using this product
  → I should get a typical journal entry containing a line with the "4458 waiting vat" account configured on the tax. Nothing different from standard, until there.
- Now pay the invoice and reconcile it to the invoice:
  → I should get a typical journal entry containing the expected payment lines, PLUS an additional 2 lines transferring the VAT amount from 4458 waiting tax to 4457 due tax. (Or possibly a different journal entry with these two lines).

Can I reproduce this scenario with this module?
I believe for this to work we only need ONE additional field to link two accounts (4458→4457), but nothing more.

Additional questions:
Why do we need a "payment on tax" on the invoice? I think this decision depends on the products (the configured tax), not really on the invoice, and in french accounting I think we are supposed to never mix the two cases on a single invoice.
and why do we need so many additional fields, on the fiscal position, the journal, the move line, etc.. ?

account_vat_on_payment/account_fiscal_position_view.xml
eLBati, line 26: As XML files can draw the odoo user interface, they should be treated as source code at the level of python files, I think.
I mean, if someone wanted to make the copyright notice explicit within XML files, I would not prevent it.

eLBati:
Thanks for your feedback @ccomb 
Try the following steps
- create demo database
- install account_vat_on_payment and account_accountant
- do not configure accounting data
- create an account.tax '20 %' and use 'X1112 Input VAT - (test)' as 'Invoice Tax Account'
- open 'Input VAT - (shadow)' account, rename to 'waiting input VAT' and set 'X110 Current Liabilities - (test)' as parent
- open 'Product Sales - (test)' account and set 'X2001 Product Sales - (test)' (itself) in 'Shadow Account for VAT on payment' field
- rename 'Bank Journal - (shadow)' journal to 'Miscellaneous Journal'
- create an invoice with 'vat on payment' and 1 line with unit price 100 and 20% VAT
- pay it with 'register payment' on the invoice

> Why do we need a "payment on tax" on the invoice?

Because in Italy, for example, some invoices can be under 'vat on payment' and others can be outside. It depends on the partner. This is why the fiscal position is used: it sets the 'vat on pyament' field on the invoice.

> I think we are supposed to never mix the two cases on a single invoice

This is correct

> and why do we need so many additional fields, on the fiscal position, the journal, the move line, etc.. ?

On the journal, the field is used for the secondary entry, see 'extra journal' at https://docs.google.com/spreadsheet/ccc?key=0Aodwq17jxF4edDJaZ2dOQkVEN0hodEtfRmpVdlg2Vnc#gid=10
On the move line, the fields are mainly used for the computation and for the links between standard and extra lines

ccomb:
Thanks for your explanations, but I used a copy of my production db configured exactly like you said (I've also set the waiting account as a shadow of the due vat account), and I've tried different other configurations but I couldn't reproduce the expected moves.
- on the income account, the balance is OK, but I have 2 unuseful credit + debit move lines
- on the customer account, everything is OK
- on the bank account, everything is OK
- on the due vat account, I get 3 credits with the vat amount (instead of just 1 credit)
- on the waiting vat account, I get 2 debits and 1 credit with the vat amount (instead of 1 debit + 1 credit)

I suppose something is still wrong in my configuration, related to the vat account but anyway, regarding the way this module works, it looks like even if the configuration and the balance eventually was ok, in the end I would still have unwanted moves. And I find the configuration very hard to understand.

Ideally, I think a "vat on payment" should be handled at lowest level, _during reconciliation_, so that we wouldn't need to depend on the voucher at all, and we wouldn't need to click on the "register payment" button of the invoice.

Here is my functionnal review summary (Lorenzo please tell again if I'm wrong):

Good points:
- it's useful and works today for some real cases (has been used for a customer)
- it's flexible, we can define 'vat on payment' on several places (customer, invoice, ...)
- it supports partial payments (I've not tested)
- it supports foreign currencies (I've not tested)
- Once configured it's simple to use for the final customer

Bad points:
- It creates unwanted additional moves (though supposed to be correctly balanced)
- It is based on the voucher code and only works if you click on the payment button of the invoice, not if you register a payment separately then reconcile with the invoice.
- It's too complex to understand and configure
- it adds too many fields in the database

So in the end I'm +0 for merging into OCA. Maybe we need another functional review or opinion.

eLBati:
> Thanks for your explanations, but I used a copy of my production db configured exactly like you said (I've also set the waiting account as a shadow of the due vat account),

Please note I suggested to use a demo DB because the account_vat_on_payment module adds some demo data (account.account and account.journal) and the suggested steps were based on that demo data.

> on the income account, the balance is OK, but I have 2 unuseful credit + debit move lines

True. The 2 additional move lines should be avoided in the French case (and in the Italian one too).
If the OCA thinks this module can be useful, we can extend the module in order to handle that

> on the due vat account, I get 3 credits with the vat amount (instead of just 1 credit)
> on the waiting vat account, I get 2 debits and 1 credit with the vat amount (instead of 1 debit + 1 credit)

This should not happen.
See my journal entries generated by the payment

bank_entry
extra_entry

my account.tax

20_tax

and my accounts

sales_account
vat_account

> Ideally, I think a "vat on payment" should be handled at lowest level, during reconciliation, so that we wouldn't need to depend on the voucher at all, and we wouldn't need to click on the "register payment" button of the invoice

Agree. While designing the module, I tried the reconciliation based way, without success. But I designed it some year ago and now, possibly with the help of some OCA experts, maybe for the future versions of odoo, I hope we'll be able to refactor the module and use the reconciliation based approach.

> It is based on the voucher code and only works if you click on the payment button of the invoice, not if you register a payment separately then reconcile with the invoice

In general, it works with vouchers and [that's bad](https://www.youtube.com/watch?v=r_pqnsKWlpc) :grin:
but it's the only solution we have for v7 at the moment

gkliska:
We are using this 6.1 module 
https://github.com/gkliska/openerp_storno_accounting/tree/master/account_tax_payment
Maybe it can help as a starting point.

account_vat_on_payment/account_invoice.py
```diff
@@ -0,0 +1,104 @@
...
+                    if (
+                        account.type != 'receivable'
+                        and account.type != 'payable'
+                    ):
+                        if not account.vat_on_payment_related_account_id:
+                            raise orm.except_orm(
+                                _('Error'),
+                                _('''The invoice is \'VAT on payment\' but
+                                account %s does not have a related shadow
+                                account''')
```
bwrsandman: Careful, this renders as:

```
The invoice is 'VAT on payment' but
                            tax code %s does not have a related shadow
                            tax code
```

account_vat_on_payment/account_invoice.py
```diff
@@ -0,0 +1,104 @@
...
+        tax_code_pool = self.pool.get('account.tax.code')
+        new_move_lines = []
+        for line_tup in move_lines:
+            if invoice_browse.vat_on_payment:
+                if line_tup[2].get('account_id', False):
+                    account = acc_pool.browse(
+                        cr, uid, line_tup[2]['account_id'])
+                    if (
+                        account.type != 'receivable'
+                        and account.type != 'payable'
```
bwrsandman: can be simplified as
`if account.type not in ['receivable', 'payable']:`

account_vat_on_payment/account_invoice.py
```diff
@@ -0,0 +1,104 @@
...
+        account and tax code related records
+        """
+        move_lines = super(account_invoice, self).finalize_invoice_move_lines(
+            cr, uid, invoice_browse, move_lines)
+        acc_pool = self.pool.get('account.account')
+        tax_code_pool = self.pool.get('account.tax.code')
+        new_move_lines = []
+        for line_tup in move_lines:
+            if invoice_browse.vat_on_payment:
+                if line_tup[2].get('account_id', False):
```
bwrsandman: At this point, you're in three levels of indentation, this is a clue to start factoring your code into sub functions.

account_vat_on_payment/account_invoice.py
```diff
@@ -0,0 +1,104 @@
...
+                    line_tup[2]['real_tax_code_id'] = line_tup[
+                        2]['tax_code_id']
+                    line_tup[2]['tax_code_id'] = (
+                        tax_code.vat_on_payment_related_tax_code_id.id)
+            new_move_lines.append(line_tup)
+        return new_move_lines
+
+    def onchange_partner_id(
+            self, cr, uid, ids, type, partner_id, date_invoice=False,
+            payment_term=False, partner_bank_id=False, company_id=False):
```
bwrsandman, line 91: Missing context

account_vat_on_payment/account_invoice.py
```diff
@@ -0,0 +1,104 @@
...
+        res = super(account_invoice, self).onchange_partner_id(
+            cr, uid, ids, type, partner_id, date_invoice, payment_term,
+            partner_bank_id, company_id)
+        # default value for VAT on Payment is changed every time the
+        # customer/supplier is changed
+        partner_obj = self.pool.get("res.partner")
+        if partner_id:
+            p = partner_obj.browse(cr, uid, partner_id)
```
bwrsandman: Missing context propagation

account_vat_on_payment/account_invoice_view.xml
```diff
@@ -0,0 +1,27 @@
...
+     <record id="invoice_form" model="ir.ui.view">
+            <field name="name">account.invoice.form</field>
+            <field name="model">account.invoice</field>
+            <field name="inherit_id" ref="account.invoice_form"/>
+            <field name="arch" type="xml">
+                <field name="payment_term" position="after">
+                    <field name="vat_on_payment"/>
+                </field>
+            </field>
+        </record>
```
bwrsandman: closing tag should be inline with opening tags

account_vat_on_payment/account_invoice_view.xml
```diff
@@ -0,0 +1,27 @@
...
+     <record id="invoice_supplier_form" model="ir.ui.view">
+            <field name="name">account.invoice.supplier.form</field>
+            <field name="model">account.invoice</field>
+            <field name="inherit_id" ref="account.invoice_supplier_form"/>
+            <field name="arch" type="xml">
+                <field name="date_due" position="after">
+                    <field name="vat_on_payment"/>
+                </field>
+            </field>
+        </record>
```
bwrsandman: closing tag should be inline with opening tags

account_vat_on_payment/account_invoice_view.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
+
+     <record id="invoice_form" model="ir.ui.view">
+            <field name="name">account.invoice.form</field>
```
bwrsandman: inconsistent indentation

account_vat_on_payment/account_invoice_view.xml
```diff
@@ -0,0 +1,27 @@
...
+            <field name="model">account.invoice</field>
+            <field name="inherit_id" ref="account.invoice_form"/>
+            <field name="arch" type="xml">
+                <field name="payment_term" position="after">
+                    <field name="vat_on_payment"/>
+                </field>
+            </field>
+        </record>
+     <record id="invoice_supplier_form" model="ir.ui.view">
+            <field name="name">account.invoice.supplier.form</field>
```
bwrsandman: Inconsistent indentation

account_vat_on_payment/account_voucher.py
```diff
@@ -0,0 +1,341 @@
...
+                    if (
+                        line.move_line_id and line.move_line_id.invoice
+                        and line.move_line_id.invoice.vat_on_payment
+                    ):
+                        vat_on_p += 1
+            if vat_on_p and vat_on_p != valid_lines:
+                raise orm.except_orm(
+                    _('Error'),
+                    _("""Can't handle VAT on payment if not every invoice
+                    is on a VAT on payment treatment"""))
```
bwrsandman: Careful with the spaces here

account_vat_on_payment/account_voucher.py
```diff
@@ -0,0 +1,341 @@
...
+                journal_pool.write(
+                    cr, uid, voucher.journal_id.id, {'entry_posted': False})
+            res = super(account_voucher, self).action_move_line_create(
+                cr, uid, [voucher.id], context)
+            # because 'move_id' has been updated by 'action_move_line_create'
+            voucher.refresh()
+            if entry_posted:
+                journal_pool.write(
+                    cr, uid, voucher.journal_id.id, {'entry_posted': True})
+            if self.is_vat_on_payment(voucher):
```
bwrsandman, line 315: You might want to factor what is in this if statement into another function

account_vat_on_payment/company.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class res_company(orm.Model):
+    _inherit = "res.company"
+    _columns = {
+        'vat_on_payment': fields.boolean('VAT on payment treatment'),
+        }
+
+
+class account_config_settings(orm.TransientModel):
```
bwrsandman: One model/class per file

account_vat_on_payment/account_invoice.py
eLBati: In this case, I can't find a way... Any suggestion?

account_vat_on_payment/account_invoice.py
eLBati, line 91: Do you suggest to add it even if onchange_partner_id of `account.invoice` does not have it?

account_vat_on_payment/account_invoice.py
bwrsandman: It's hard to tell what is being done in that block (which of itself is a problem due to the monolith of a function).

I suggest looking at that block, seeing what you're trying to do and what objects are being affected and what should be returned. Move that block to a function, pass what you need (`invoice_browse.vat_on_payment`, `line_tup[2]`, etc) and return `line_tup` or raise an exception.

The problem is that it's taking me way too long just to understand the data being manipulated. Imagine what it would be like for a new dev to try and maintain this part. It should be immediately apparent what the code is doing with regards to variable names and function names, not even with comments.

account_vat_on_payment/account_invoice.py
bwrsandman, line 91: Absolutely, otherwise all your manipulations from then on will lose i18n.

It seems onchange functions never have context, if that's the case, you can reclaim it.

Put this at the top of you onchange:

``` python
    if context is None:
        context = self.pool['res.users'].context_get(cr, uid)
```

eLBati:
Thanks @bwrsandman I followed your suggestions

account_vat_on_payment/account_account.py
```diff
@@ -0,0 +1,34 @@
...
+from openerp.osv import orm, fields
+
+
+class account_account(orm.Model):
+    _inherit = "account.account"
+    _columns = {
+        'vat_on_payment_related_account_id': fields.many2one(
+            'account.account', 'Shadow Account for VAT on payment',
+            help='''Related account used for real registrations on a
+            VAT on payment basis. Set the shadow account here'''),
```
bwrsandman: There are too many spaces between a and VAT here. You may want to do:

``` python
            help='Related account used for real registrations on a '
                 'VAT on payment basis. Set the shadow account here'
```

account_vat_on_payment/account_demo.xml
```diff
@@ -0,0 +1,56 @@
...
+            <field name="code">IVAS</field>
+            <field name="name">Input VAT - (shadow)</field>
+            <field ref="scoa" name="parent_id"/>
+            <field name="type">other</field>
+            <field name="user_type" ref="account.data_account_type_liability"/>
+        </record>
+        
+        <!-- real COA -->
+        <record id="account.a_sale" model="account.account">
+            <field name="vat_on_payment_related_account_id" ref="pss"></field>
```
bwrsandman: XML tags of this style can be simplifed like this:

``` xml
<field name="vat_on_payment_related_account_id" ref="pss"/>
```

account_vat_on_payment/account_demo.xml
```diff
@@ -0,0 +1,56 @@
...
+        </record>
+        <record id="account.iva" model="account.account">
+            <field name="vat_on_payment_related_account_id" ref="ivas"></field>
```
bwrsandman: Same as https://github.com/OCA/account-payment/pull/2/files#r19033557

account_vat_on_payment/account_demo.xml
```diff
@@ -0,0 +1,56 @@
...
+        </record>
+        
+        <!-- Journals -->
+        <record id="shadow_bank_journal" model="account.journal">
+            <field name="name">Bank Journal - (shadow)</field>
+            <field name="code">SBNK</field>
+            <field name="type">general</field>
+        </record>
+        <record id="account.bank_journal" model="account.journal">
+            <field name="vat_on_payment_related_journal_id" ref="shadow_bank_journal"></field>
```
bwrsandman: Same as https://github.com/OCA/account-payment/pull/2/files#r19033557

account_vat_on_payment/account_demo.xml
```diff
@@ -0,0 +1,56 @@
...
+        </record>
+        <record id="account.bank_journal_usd" model="account.journal">
+            <field name="vat_on_payment_related_journal_id" ref="shadow_bank_journal"></field>
```
bwrsandman: Same as https://github.com/OCA/account-payment/pull/2/files#r19033557

account_vat_on_payment/account_invoice.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def _set_vat_on_payment_account(self, cr, uid, line_tuple, context=None):
+        acc_pool = self.pool.get('account.account')
+        account = acc_pool.browse(
+            cr, uid, line_tuple[2]['account_id'], context=context)
+        if account.type not in ['receivable', 'payable']:
+            if not account.vat_on_payment_related_account_id:
+                raise orm.except_orm(
+                    _('Error'),
+                    _("The invoice is \'VAT on payment\' but "
```
bwrsandman: `\'` is not necessary here

account_vat_on_payment/account_invoice.py
```diff
@@ -0,0 +1,113 @@
...
+        return line_tuple
+
+    def _set_vat_on_payment_tax_code(self, cr, uid, line_tuple, context=None):
+        tax_code_pool = self.pool.get('account.tax.code')
+        tax_code = tax_code_pool.browse(
+            cr, uid, line_tuple[2]['tax_code_id'], context=context)
+        if not tax_code.vat_on_payment_related_tax_code_id:
+            raise orm.except_orm(
+                _('Error'),
+                _('''The invoice is \'VAT on payment\' but
```
bwrsandman: Same as https://github.com/OCA/account-payment/pull/2/files#r19033936

account_vat_on_payment/account_journal.py
```diff
@@ -0,0 +1,34 @@
...
+from openerp.osv import orm, fields
+
+
+class account_journal(orm.Model):
+    _inherit = "account.journal"
+    _columns = {
+        'vat_on_payment_related_journal_id': fields.many2one(
+            'account.journal', 'Shadow Journal for VAT on payment',
+            help='''Related journal used for shadow registrations on a
+            VAT on payment basis. Set the shadow journal here'''),
```
bwrsandman: Same as https://github.com/OCA/account-payment/pull/2/files#r19033479

account_vat_on_payment/account_tax_code.py
```diff
@@ -0,0 +1,34 @@
...
+from openerp.osv import orm, fields
+
+
+class account_tax_code(orm.Model):
+    _inherit = "account.tax.code"
+    _columns = {
+        'vat_on_payment_related_tax_code_id': fields.many2one(
+            'account.tax.code', 'Shadow Tax code for VAT on payment',
+            help='''Related tax code used for real registrations on a
+            VAT on payment basis. Set the shadow tax code here'''),
```
bwrsandman: Same as https://github.com/OCA/account-payment/pull/2/files#r19033479

account_vat_on_payment/account_voucher.py
```diff
@@ -0,0 +1,346 @@
...
+        if voucher.type in ('payment', 'receipt'):
+            for line in voucher.line_ids:
+                if line.amount:
+                    valid_lines += 1
+                    if (
+                        line.move_line_id and line.move_line_id.invoice
+                        and line.move_line_id.invoice.vat_on_payment
+                    ):
+                        vat_on_p += 1
+            if vat_on_p and vat_on_p != valid_lines:
```
bwrsandman, line 48: Couldn't you just raise the exception here:

``` python
if not (
    line.move_line_id and line.move_line_id.invoice
    and line.move_line_id.invoice.vat_on_payment
):
```

Then do away with `vat_on_p` and return `valid_lines`

account_vat_on_payment/account_voucher.py
```diff
@@ -0,0 +1,346 @@
...
+            # That means: use the invoice's total
+            # (as we are in 'full reconcile' case)
+            allocated_amount = amounts_by_invoice[invoice.id]['allocated']
+        else:
+            allocated_amount = (
+                amounts_by_invoice[invoice.id]['allocated']
+                +
+                amounts_by_invoice[invoice.id]['write-off']
+                )
+        new_line_amount = currency_obj.round(
```
bwrsandman, line 81: This could be simpler with like this:

``` python
allocated_amount = amounts_by_invoice[invoice.id]['allocated']
if not (
    voucher.exclude_write_off
    and voucher.payment_option == 'with_writeoff'
):
    allocated_amount += amounts_by_invoice[invoice.id]['write-off']
```

account_vat_on_payment/account_voucher.py
```diff
@@ -0,0 +1,346 @@
...
+                # avoid including write-off if set in
+                # voucher.
+                allocated_amount = amounts_by_invoice[
+                    invoice.id]['allocated_currency']
+            else:
+                allocated_amount = (
+                    amounts_by_invoice[invoice.id]['allocated_currency']
+                    +
+                    amounts_by_invoice[invoice.id]['currency-write-off']
+                )
```
bwrsandman: Seems like there is copied code from `_compute_new_line_amount` that should not be the case, please factor that into a function.

account_vat_on_payment/account_voucher.py
```diff
@@ -0,0 +1,346 @@
...
+        if new_line_amount_curr:
+            vals['amount_currency'] = new_line_amount_curr
+            vals['currency_id'] = foreign_curr_id
+        if inv_move_line.tax_code_id:
+            if not inv_move_line.real_tax_code_id:
+                raise orm.except_orm(
+                    _('Error'),
+                    _("""We are on a VAT on payment
+                    treatment but move line %s does not
+                    have a related real tax code""")
```
bwrsandman: Careful about the spaces between your words on new lines

account_vat_on_payment/account_voucher.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def _create_vat_on_payment_move(self, cr, uid, voucher, context=None):
+        move_line_pool = self.pool.get('account.move.line')
+        move_pool = self.pool.get('account.move')
+        inv_pool = self.pool.get('account.invoice')
+        if not voucher.journal_id.vat_on_payment_related_journal_id:
+            raise orm.except_orm(
+                _('Error'),
+                _("""We are on a VAT on payment treatment
+                but journal %s does not have a related shadow
```
bwrsandman: Same as https://github.com/OCA/account-payment/pull/2/files#r19035366

account_vat_on_payment/account_voucher.py
eLBati, line 48: The purpose of the check `if vat_on_p and vat_on_p != valid_lines` is to guarantee that, if the voucher contains at least 1 'vat on payment' invoice, then every invoice of the voucher must be 'vat on payment'

eLBati:
Thanks @bwrsandman I made the changes

account_vat_on_payment/account_invoice.py
```diff
@@ -0,0 +1,113 @@
...
+    def _set_vat_on_payment_tax_code(self, cr, uid, line_tuple, context=None):
+        tax_code_pool = self.pool.get('account.tax.code')
+        tax_code = tax_code_pool.browse(
+            cr, uid, line_tuple[2]['tax_code_id'], context=context)
+        if not tax_code.vat_on_payment_related_tax_code_id:
+            raise orm.except_orm(
+                _('Error'),
+                _('''The invoice is 'VAT on payment' but
+                tax code %s does not have a related shadow
+                tax code''')
```
bwrsandman: Careful about the N spaces at the start of each line.

account_vat_on_payment/account_invoice.py
bwrsandman, line 91: :+1:

account_vat_on_payment/account_voucher.py
bwrsandman, line 315: :+1:

bwrsandman:
:+1: 
Looks pretty good to me.

max3903:
The pot file is missing

:-1:

tafaRU:
@eLBati, https://github.com/eLBati/account-payment/pull/1 adds the pot file.

eLBati:
thanks @max3903 and @tafaRU, pot file added

tafaRU:
@eLBati  please also consider https://github.com/eLBati/account-payment/pull/2

bwrsandman:
:+1: Code review

max3903:
:+1:

foutoucour:
:+1:

guewen:
Green, several thumbs up, tests cases and this is the oldest PR on the OCA repositories. The reviewers/reviewee works has been proficient here! Let's merge it.

PR #5, opened 2014-07-28
7.0 metafiles

pedrobaeza:


oihane:
PEP8 fails but LGTM :+1:

yvaucher:
mx.Datetime in paydays module should be replaced by datetime.datetime

This can be done in another PR

:+1:

PR #12, opened 2014-10-24
7.0 account voucher default writeoff

joaoalf:
This PR add the new module account_voucher_default_writeoff that changes thes payment mode to with writeoff when there's a writeoff amount.

### DIFF
--- a/account_voucher_default_writeoff/__init__.py
+++ b/account_voucher_default_writeoff/__init__.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Thanks for the module, but I think this is a litle functionality that doesn't deserve a module, because it's only one-click distance for user.

What the others think?

joaoalf:
Actually, this one click is a source of constant error in many of our clients. With the writeoff option there's the account field that became required so the user cannot finish the voucher without declaring the proper account for the difference value.

pedrobaeza:
It's a good motivation indeed. Let's see what others think.

joaoalf:
Sure. Thanks for your comments.

eLBati:
LGTM :+1:

max3903:
:+1:

PR #13, opened 2014-10-28
7.0 account voucher no auto lines

joaoalf:
This module changes the voucher behavior to allow users to manually select the lines that will be payed.

### DIFF
--- a/account_voucher_no_auto_lines/__init__.py
+++ b/account_voucher_no_auto_lines/__init__.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+###############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    This module copyright (C) 2014 Savoir-faire Linux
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Good solution. :+1: 

Tests will be much appreciated, at even more with this hacky solution.

joaoalf:
Thanks for the comments @pedrobaeza 

What kind of test do you have in mind? I ask because since the onchange methods are made to affect the client behavior and not the server, to test if I'm still able to create a voucher won't prove much and to test if the onchange returns the right struct sounds too trivial. It would be nice to have real client side tests to validate the onchange behaviour but so far the client side tests features of Odoo are too obscure to be trusted.

pedrobaeza:
Well, although a simple test like create an account invoice, then an account voucher, and check if the amount is 0, this can very useful when migrating this to future versions. YAML tests trigger onchange automatically, so there's no problem with that.

PR #14, opened 2014-10-29
Update travis, Readme and remove .coveragerc

bwrsandman:

PR #15, opened 2014-11-06
[8.0] Update travis, Readme and remove .coveragerc

bwrsandman:
Forward port of #14


pedrobaeza:
:+1:

yvaucher:
:+1:

sbidoul:
If thread [1] reaches a positive conclusion, the 8.0 port of account-payment should not be done under OCA umbrella. The thread is currently at three +1 votes and one -0 vote.

Abandoning a repo may be perceived as a bit radical, but in this case, the functionality is close to being completely covered by OCA/banking 8.0 and OCA may not have the resource to maintain duplicate functionality.

So I suggest suspending this PR until [1] as reached a definitive conclusion.

[1] https://lists.launchpad.net/banking-addons-drivers/msg00120.html

PR #21, opened 2014-12-04
7.0 Added modules 'account_voucher_remit_to' and 'account_check_writing_remit_to'

JordiBForgeFlow:
In countries where payments are made by checks the account vouchers are used, and it often happens that the addresses that the checks should be sent to is different to the address of the supplier.

The new modules included in this PR add a new address type on the partner 'Remit to'.
During the creation of a Voucher, once the partner is entered the partner's remit-to is fetched and added to the voucher.

See a blog post that explains it:
http://goo.gl/KRbdJW

Regards,
Jordi Ballester


JordiBForgeFlow:
I changed the logic, so that a partner address type 'remit-to' is used instead of partner relationships.

JordiBForgeFlow:
I'm replacing this with https://github.com/OCA/account-payment/pull/53

PR #25, opened 2014-12-22
[add] account_due_list

scigghia:


sbidoul:
Hello Andrea,

Thanks for this PR!

Can you move this PR to the OCA/bank-payment repository? All payment modules are living there from 8.0 onwards (except the voucher-based modules which will probably stay here).

scigghia:
sure!
(i've found account_due_list only here)

scigghia:
https://github.com/OCA/bank-payment/pull/84

PR #28, opened 2015-01-26
8.0 port account_voucher_supplier_invoice_number

JehoG:
I would like to change all the file for the V8, but when _get_supplier_invoice_number is updated to V8, the field stop being set for an unknown reason.


account_voucher_supplier_invoice_number/voucher.py
```diff
@@ -40,10 +40,10 @@ def _get_supplier_invoice_number(self, cr, uid, ids, name, args, context=None):
     
     _columns = {
```
pedrobaeza: If you put models, you should declare variables in new style.

account_voucher_supplier_invoice_number/voucher.py
```diff
@@ -40,10 +40,10 @@ def _get_supplier_invoice_number(self, cr, uid, ids, name, args, context=None):
...
         'supplier_invoice_number': fields.function(_get_supplier_invoice_number,
-            type='char', size=64, string="Supplier Invoice Number"),
+            type='char', size=64, string=_("Supplier Invoice Number")),
         }
 
-class voucher(orm.Model):
+class voucher(models.Model):
     _inherit = 'account.voucher'
     
     def recompute_voucher_lines(self, cr, uid, ids, partner_id, journal_id, price,
```
pedrobaeza: Old API, but model declaration in new API...

JehoG: Ya, I know, but when I change it all like there (only  _get_supplier_invoice_number):
https://github.com/JehoG/account-payment/blob/Example/account_voucher_supplier_invoice_number/voucher.py
it's just leaving the field blank (in the form view fo a supplier payment under 'supplier invoice number'. But if I check at the value of res just before the line 40, I got what it's suppose to be.
So I got told to push it only with those minor changes if it's working.
You see something that I missed in this example?

pedrobaeza: You are mixing new and old API. If you do all in new API, you won't find these problems.

JehoG:
Here's the file with the V8 api. But wasnt working neither. Do you see what is wrong?

account_voucher_supplier_invoice_number/voucher.py
```diff
@@ -19,47 +19,55 @@
...
-from tools.translate import _
+from openerp import api, fields, models, _
 
-class voucher_line(orm.Model):
+class voucher_line(models.Model):
     _inherit = 'account.voucher.line'
     
-    def get_suppl_inv_num(self, cr, uid, move_line_id, context=None):
-        move_line = self.pool.get('account.move.line').browse(cr, uid, move_line_id, context)
+    def get_suppl_inv_num(self, move_line_id):
```
pedrobaeza, line 27: Why two methods?

account_voucher_supplier_invoice_number/voucher.py
```diff
@@ -19,47 +19,55 @@
...
+    def _get_supplier_invoice_number(self):
         res={}
-        for line in self.browse(cr, uid, ids, context):
+        for line in self:
             res[line.id] = ''
             if line.move_line_id:
-                res[line.id] = self.get_suppl_inv_num(cr, uid,
-                    line.move_line_id.id, context=context)
+                res[line.id] = self.get_suppl_inv_num(line.move_line_id.id)
+        print res
```
pedrobaeza, line 44: New API you don't have to return a dictionary, but assign directly the value: `self.supplier_invoice_number = ...`

scigghia:
https://github.com/OCA/OCB/pull/128

is this usefull ? whit this PR we can see supplier invoice number by default in account voucher

JehoG:
True, this PR seems finally useless with your PR.
Thanks for the hint and thanks to pedrobaeza  for your help.
Closing the PR.

PR #35, opened 2015-05-15
[ADD] account_payment_extension: Migration scripts to bank-payment modules

pedrobaeza:
# Migration scripts for account_payment_extension v7 set of modules

This module includes the needed migration script for making a smooth transition
from v7 _account_payment_extension_ module to the set of v8 bank-payment
modules.

This module doesn't provide any functionality at user level.
# Installation

You need OpenUpgrade for using these scripts. Put this module in a path
included on OpenUpgrade addons path. You need also to have accesible the
repository bank-payment from https://github.com/OCA/bank-payment.

Start the OpenUpgrade migration and the conversion will be automatically done,
and this module will be automatically uninstall.


angelmoya:
:+1:

account_payment_extension/migrations/8.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,121 @@
...
+                # Create the payment mode
+                openupgrade.logged_query(
+                    cr,
+                    """
+                    INSERT INTO payment_mode
+                    (%s, name, active, journal, bank_id, company_id)
+                    VALUES (%%s, %%s, %%s, %%s, %%s)
+                    RETURNING id;""" % openupgrade.get_legacy_name('type'),
+                    (payment_type[0], payment_type[1], payment_type[2],
+                     journal_id, bank_id, company_id))
```
cubells: This pull works like a charm but I've had to change this insert code in order to be able to run it. For me works this:

```
                # Create the payment mode
                cr.execute(
                    """
                    INSERT INTO payment_mode
                    (id, name, active, journal, bank_id, company_id)
                    VALUES (%s, '%s', %s, %s, %s, %s)
                    RETURNING id;""" %
                    (payment_type[0], payment_type[1], payment_type[2],
                     journal_id, bank_id, company_id,))
```

pedrobaeza: What is the error? I think this is because and old PG version, and with that SQL, you can overwrite some id values.

cubells: I've got two errors. One is easy:

```
openerp.sql_db: bad query: 
                    INSERT INTO payment_mode
                    (openupgrade_legacy_8_0_type, name, active, journal, bank_id, company_id)
                    VALUES (%s, %s, %s, %s, %s)
                    RETURNING id;
Traceback (most recent call last):
  File "/home/cubells/projectes/git/odoo80/8.0/server/openerp/sql_db.py", line 234, in execute
    res = self._obj.execute(query, params)
TypeError: not all arguments converted during string formatting
```

The second is the error that made me to change the insert code:

```
2016-02-10 05:06:17,078 15371 ERROR openerp.sql_db: Programming error: column "openupgrade_legacy_8_0_type" of relation "payment_mode" does not exist
LINE 3:                     (openupgrade_legacy_8_0_type, name, acti...
                             ^
, in query 
                    INSERT INTO payment_mode
                    (openupgrade_legacy_8_0_type, name, active, journal, bank_id, company_id)
                    VALUES (%s, %s, %s, %s, %s, %s)
                    RETURNING id;
```

pedrobaeza:
@cubells, the correct SQL statement has been fixed in https://github.com/pedrobaeza/account-payment/commit/d56f73fc3ebb37dc0d722553d133fcbcc816dd12

pedrobaeza:
As all comments have been honoured and the Travis problem is unrelated, I merge this one to have it finally on the repo.

sebalix:
Arg, too late... I found a bug and was not able to find your branch anymore ;)
See https://github.com/OCA/account-payment/pull/69

PR #42, opened 2015-08-21
[FIX] account_payment_term_multi_day: Use old API for method compute due to argument incompatibility

pedrobaeza:
This method is called from multiple points of the code following different techniques:
- https://github.com/odoo/odoo/blob/8.0/addons/account/account_move_line.py#L713
- https://github.com/odoo/odoo/blob/8.0/addons/account/account_invoice.py#L508

And the non-standarization of the name of the arguments makes the new API wrapper to fail, so the only solution is to provide it as old API method.

### DIFF
--- a/account_payment_term_multi_day/models/account_payment_term.py
+++ b/account_payment_term_multi_day/models/account_payment_term.py
@@ -23,12 +23,16 @@ import calendar
 class AccountPaymentTerm(models.Model):
     _inherit = "account.payment.term"
 
-    @api.multi
-    def compute(self, value, date_ref=False):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
cc @antespi @amoya-dx @omar7r

antespi:
:+1: code review

gurneyalex:
:+1: no test.

I did a PR for a similar case which was accepted in odoo 8.0 a while ago (https://github.com/odoo/odoo/pull/7911). Maybe you could do the same

gurneyalex:
fast tracking the fix.

PR #44, opened 2015-09-03
ADD account_payment_order_to_voucher module

eLBati:
# 
# Payment order to voucher

This module allows to generate supplier vouchers from payment order,
based on payment order data
# Usage

For 'done' payment orders, click on 'Create vouchers' button

### DIFF
--- a/account_payment_order_to_voucher/models/payment_order.py
+++ b/account_payment_order_to_voucher/models/payment_order.py
@@ -1,0 +1,98 @@
+# -*- coding: utf-8 -*-
+#
+#   See __openerp__.py about license
+#
+
+from openerp import models, fields, api, _
+from openerp.exceptions import Warning
+
+
+class PaymentOrder(models.Model):
+    _inherit = 'payment.order'
+    voucher_ids = fields.Many2many(
+        'account.voucher', string='Vouchers', readonly=True)
+
+    def get_lines_by_partner(self, order):
+        lines_by_partner = {}
+        if order.voucher_ids:
+            raise Warning(
+                _("Payment order %s already has vouchers")
+                % order.reference)
+        if order.state != 'done':
+            raise Warning(
+                _("Payment order %s is not in 'done' state")
+                % order.reference)
+        for line in order.line_ids:
+            if line.partner_id.id not in lines_by_partner:
+                lines_by_partner[line.partner_id.id] = self.env['payment.line']
+            lines_by_partner[line.partner_id.id] |= line
+        return lines_by_partner
+
+    def _compute_lines_total(self, payment_lines):
+        return sum(payment_lines.mapped('amount_currency'))
+
+    def _get_currency_id(self, payment_lines):
+        currency_ids = payment_lines.mapped('currency').ids
+        if len(currency_ids) > 1:
+            raise Warning(
+                _("Every order lines must have the same currency"))
+        return currency_ids[0]
+
+    def _build_voucher_header(self, payment_lines):
+        total = self._compute_lines_total(payment_lines)
+
+        # every line has the same order and partner
+        order = payment_lines[0].order_id
+        partner = payment_lines[0].partner_id
+
+        currency_id = self._get_currency_id(payment_lines)
+        voucher_vals = {
+            'type': 'payment',
+            'name': order.reference,
+            'partner_id': partner.id,
+            'journal_id': order.mode.journal.id,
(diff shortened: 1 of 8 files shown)


account_payment_order_to_voucher/models/payment_order.py
```diff
@@ -0,0 +1,102 @@
...
+                _("Payment order %s is not in 'done' state")
+                % order.reference)
+        for line in order.line_ids:
+            if line.partner_id.id not in lines_by_partner:
+                lines_by_partner[line.partner_id.id] = []
+            lines_by_partner[line.partner_id.id].append(line)
+        return lines_by_partner
+
+    def _compute_lines_total(self, payment_lines):
+        total = 0.0
```
pedrobaeza: `return sum(payment_lines.mapped('amount_currency')`

account_payment_order_to_voucher/models/payment_order.py
```diff
@@ -0,0 +1,102 @@
...
+        for line in payment_lines:
+            total += line.amount_currency
+        return total
+
+    def _get_currency_id(self, payment_lines):
+        currency_ids = [l.currency.id for l in payment_lines]
```
pedrobaeza: `currencies = payment_lines.mapped('currency')`

account_payment_order_to_voucher/views/payment_order_view.xml
```diff
@@ -0,0 +1,23 @@
...
+    <data>
+        <record id="view_payment_order_form" model="ir.ui.view">
+            <field name="name">payment.order.form</field>
+            <field name="model">payment.order</field>
+            <field name="inherit_id" ref="account_payment.view_payment_order_form"></field>
+            <field name="arch" type="xml">
+                <div class=" oe_right oe_button_box" position="after">
+                    <div class=" oe_right oe_button_box">
+                        <button class="oe_inline oe_stat_button oe_right" name="generate_vouchers" string="Create vouchers"
+                               type="object" attrs="{'invisible':[('state','!=','done')]}" icon="fa-pencil-square-o" widget="statinfo"/>
```
pedrobaeza: Limit visibility here also for vouchers_ids != False and you save the warning in the code

pedrobaeza: ...or program it to return a view with the voucher instead.

pedrobaeza:
It would be good to have a glue module that allows this also for customers with _account_direct_debit_

account_payment_order_to_voucher/models/payment_order.py
eLBati: In this case I would keep the current, because 1 line would become 2:

```
        currencies = payment_lines.mapped('currency')
        currency_ids = currencies.mapped('id')
```

account_payment_order_to_voucher/models/payment_order.py
eLBati: actually, `payment_lines` is a list, not a recordset

account_payment_order_to_voucher/views/payment_order_view.xml
eLBati: Thanks, done

account_payment_order_to_voucher/models/payment_order.py
pedrobaeza: Well, if you need ids, then make: `currency_ids = payment_lines.mapped('currency').ids`, but I advise you to store always recordsets and only put ids in search and so on.

account_payment_order_to_voucher/models/payment_order.py
```diff
@@ -0,0 +1,102 @@
...
+            raise Warning(
+                _("Payment order %s already has vouchers")
+                % order.reference)
+        if order.state != 'done':
+            raise Warning(
+                _("Payment order %s is not in 'done' state")
+                % order.reference)
+        for line in order.line_ids:
+            if line.partner_id.id not in lines_by_partner:
+                lines_by_partner[line.partner_id.id] = []
```
pedrobaeza: You can put here `lines_by_partner[line.partner_id.id] = self.env['account.move.line']` to initialize to an empty recordset, and then add elements with `|=` operator, or directly assign the `line`variable the first time

eLBati: Thanks, done

pedrobaeza:
:+1:

archetipo:
:+1:

tafaRU:
:+1:

PR #45, opened 2015-09-10
added field amount_residual in payment_view

scigghia:
it is usefull in due list, to see every residual for each maturity, with total

### DIFF
--- a/account_due_list/account_move_line.py
+++ b/account_due_list/account_move_line.py
@@ -45,6 +45,62 @@ class AccountMoveLine(models.Model):
                                         compute='_get_invoice',
                                         string='Invoice', store=True)
 
+    maturity_residual = fields.Float(
+        compute='_maturity_residual', string="Residual Amount", store=True,
+        help="The residual amount on a receivable or payable of a journal "\
+             "entry expressed in the company currency.")
+
+    @api.multi
+    @api.depends('date_maturity')
+    def _maturity_residual(self):
+        """
+            inspired by amount_residual
+        """
+        # currency can be used in future
+        cur_obj = self.pool['res.currency']
+        for move_line in self:
+            move_line.maturity_residual= 0.0
+
+            if move_line.reconcile_id:
+                continue
+            if not move_line.account_id.reconcile:
+                #this function does not suport to be used on move lines not
+                #related to a reconcilable account
+                continue
+
+            if move_line.currency_id:
+                move_line_total = move_line.amount_currency
+                sign = move_line.amount_currency < 0 and -1 or 1
+            else:
+                move_line_total = move_line.debit - move_line.credit
+                sign = (move_line.debit - move_line.credit) < 0 and -1 or 1
+            line_total_in_company_currency =  move_line.debit - move_line.credit
+            context_unreconciled = self._context.copy()
+            if move_line.reconcile_partial_id:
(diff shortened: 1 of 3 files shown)


eLBati:
Thanks @scigghia 
adding computed fields to tree views could make loading time considerably lower. As in this case the computation is not trivial, you  should probably add a stored field

scigghia:
@eLBati amount_residual is a field in standard account.move.line and it is not stored.
so i think it is better to add a new field, what do you think about?

moylop260:
+1 with add a new field function `store=True`
Note: You will need to consider the `invoice_date` field here.

scigghia:
done!
I think dependency on date_maturity is usefull also for manual account moves

moylop260:
Please, review travis red and execute a final rebase to delete last 3 commits (pep8 fixes)

scigghia:
hi @moylop260 , travis is green

moylop260:
:+1:

giudalo:
+1:

giudalo:
:+1:

account_due_list/account_move_line.py
```diff
@@ -45,6 +45,66 @@ class AccountMoveLine(models.Model):
...
                                         string='Invoice', store=True)
 
+    maturity_residual = fields.Float(
+        compute='_maturity_residual', string="Residual Amount", store=True,
+        help="The residual amount on a receivable or payable of a journal "
+             "entry expressed in the company currency.")
+
+    @api.multi
+    @api.depends('date_maturity')
+    def _maturity_residual(self):
```
eLBati, line 58: Is it possible to call `_amount_residual` instead of partially duplicating the code?

account_due_list/account_move_line.py
```diff
@@ -45,6 +45,66 @@ class AccountMoveLine(models.Model):
                                         compute='_get_invoice',
                                         string='Invoice', store=True)
 
+    maturity_residual = fields.Float(
+        compute='_maturity_residual', string="Residual Amount", store=True,
+        help="The residual amount on a receivable or payable of a journal "
+             "entry expressed in the company currency.")
+
+    @api.multi
+    @api.depends('date_maturity')
```
eLBati: It seems to me it should also depend on other fields, like reconcile_id, account_id.reconcile, amount_currency, debit, credit, reconcile_partial_id.line_partial_ids, currency_id, company_id.currency_id

account_due_list/account_move_line.py
scigghia, line 58: sure! done

account_due_list/account_move_line.py
scigghia: added debit,credit,reconcile_id and reconcile_partial_id deps

scigghia: maybe amount_currency,currency_id and company_id.currency_id could be usefull with another field like maturity_residual_currency

eLBati: The fields I listed above are  used by `amount_residual`.
That is, for example, if you don't put `reconcile_partial_id.line_partial_ids` and if a move.line is added to `reconcile_partial_id.line_partial_ids` , `maturity_residual` is not recomputed

andrea4ever:
LGTM :+1:

giudalo:
:+1:

OpenCode:
:+1:

PR #57, opened 2016-01-14
[FIX] account_payment_order_to_voucher: bad field name (vouchers_ids) and filter in view

ozono:
account_payment.view_payment_order_form not loading due to error in inherited view: bad field name (vouchers_ids) and filter in view_payment_order_form


eLBati:
Thanks @ozono , commit https://github.com/ozono/account-payment/commit/f48022ef2dd39b515b303df04871906ffcc12b79 is not linked to any user. Can you check?

PR #60, opened 2016-02-02
[7.0] Fix unit test in account_voucher_no_auto_lines

ghost:


ghost:
Otherwise, the test fails because no currency rate is available for the period.

PR #74, opened 2016-04-05
check if currency id not changed per company, remove it from create v…

archetipo:
…alues

This PR solve an issue related by [this write method](https://github.com/odoo/odoo/commit/46721a88bff1aa0f582bb9cd635422a00c6542d3#diff-1a513ea93ae1aeca4005dcf8b810ed5aR154)
This method, if is also installed a module like [this](https://github.com/archetipo/account-payment/blob/9.0/account_due_list/account_move_line.py#L54) , generate a recalculation of all move lines, because trigger onchanege of currency_id

### DIFF
--- a/account_due_list/models/account_move_line.py
+++ b/account_due_list/models/account_move_line.py
@@ -51,7 +51,7 @@ class AccountMoveLine(models.Model):
     @api.depends('date_maturity', 'debit', 'credit', 'reconcile_id',
                  'reconcile_partial_id', 'account_id.reconcile',
                  'amount_currency', 'reconcile_partial_id.line_partial_ids',
-                 'currency_id', 'company_id.currency_id')
+                 'currency_id')
     def _maturity_residual(self):
         """
             inspired by amount_residual


pedrobaeza:
Don't you think it's better to remove that dependency instead?

account_due_list/models/res_config.py
```diff
@@ -0,0 +1,20 @@
...
+class AccountConfigSettings(models.TransientModel):
+    _inherit = 'account.config.settings'
+
+    @api.model
+    @api.returns('self', lambda value: value.id)
+    def create(self, values):
+        companyModel = self.env['res.company']
+        company = companyModel.browse(values['company_id'])
+        if values['currency_id'] == company.currency_id.id:
+            values.pop('currency_id')
```
eLBati: @archetipo could you add a comment explaining why this override is needed?

eLBati:
@pedrobaeza do you mean the dependency on `currency_id`?
It is needed by
https://github.com/OCA/account-payment/blob/8.0/account_due_list/models/account_move_line.py#L50-L61
because `amount_residual` is not stored and its computation depends on `company_id.currency_id`

pedrobaeza:
Yeah, but why it's needed? We already have currency_id in the depends list

eLBati:
`currency_id` (`account.move.line` field) and `company_id.currency_id` (`res.company` field) are both needed in the depends list of `_maturity_residual` because they are both used by [_amount_residual](https://github.com/odoo/odoo/blob/8.0/addons/account/account_move_line.py#L121), that is the method that computes `amount_residual`, used at https://github.com/OCA/account-payment/blob/8.0/account_due_list/models/account_move_line.py#L61

pedrobaeza:
Yeah, but I don't see that you need that fields. They are not used in the compute method.

eLBati:
Actually they are. See https://github.com/odoo/odoo/blob/8.0/addons/account/account_move_line.py#L159-L162

pedrobaeza:
Yeah, I know that in some moment of the computation, this field is used, but that doesn't mean you have to include as dependency, because it's not data that is related to the life of the document. I mean, if you change the company currency, you don't want that all the invoices are recomputed due to this change. In fact, you are making something similar in this PR.

That's why I say it's better to simply remove that field from the depends list.

eLBati:
Thanks to @pedrobaeza hints, I looked deeply at
https://github.com/odoo/odoo/blob/8.0/addons/account/account_move_line.py#L121-L170
and noticed `company_id.currency_id` is only used to compute `amount_residual_currency`, not `amount_residual`.

As
https://github.com/OCA/account-payment/blob/8.0/account_due_list/models/account_move_line.py#L55-L61
only uses `amount_residual`, I think we can just remove `company_id.currency_id` from `@api.depends`.

@archetipo what do you think?

PS: @pedrobaeza this PR is not bypassing computation when `company_id.currency_id` changes, is bypassing computation when `company_id.currency_id` **does not** change

archetipo:
@pedrobaeza  @eLBati 
thanks! pr updated

pedrobaeza (approved):

eLBati:
:+1:

eLBati (approved):

PR #77, opened 2016-04-11
[FIX] account_payment_order_to_voucher: Fix Travis with account_banking_payment_export

pedrobaeza:
Both modules installed conflict because the tests of this module doesn't set correctly active_model context variable.


pedrobaeza:
As the test has been fixed and this is only what we touch, I self-merge to start from a clean branch.

PR #80, opened 2016-05-11
[8.0][IMP][account_due_list] Add invoice salesperson to account.move.line for filtering and grouping

antespi:
This PR adds a stored related field `invoice_user_id` for filtering and grouping `account.move.line` by salesperson


pedrobaeza:
We also talk about another field to group/filter. Do you remember which?

antespi:
There is a new field (Salesperson) for filtering while writing in search box, and there is a group by Salesperson for grouping.
I don't remember more than those in this addon

pedrobaeza:
Thanks :+1:

eLBati:
LGTM :+1:

Issue #83, opened 2016-05-16
[8.0] [account_payment_return] Duplicated journal item


sandrafig:
**Steps to reproduce:**
- Mark as paid two sale invoices from the same partner, selecting the same bank journal.
- Go to Accounting/Customer Payment Returns and create a new one, select the bank journal and in one line add an item with the partner and the two payments. 
- Confirm the return.
- Go to Accounting/Journal Entries and open the one created by the return.

**Problem:**
- In the journal entry, there's a duplicated journal item. There are two journal items with the total amount of the two invoices.

**Expected behavior:**
- In the journal entry should be one journal item with the amount of one of the invoices, and another journal item with the amount of the other invoice, or at least one journal item with the added amount.

pedrobaeza:
cc @carlosdauden

sandrafig:
I've been trying out different scenarios and I think the module should be set to Not installable until this is fixed.

If you create a new return for two invoices of the same partner, it does not matter whether you put it in the same line or in two different lines. When you put it in two different lines, apparently everything seems OK, but those two invoices stay related (they're reconciled together) and if you try to register a payment later the invoices will not be marked as paid.

If you try to pay two invoices of the same partner, return them, and then mark them as paid again, you will see what I am talking about.

It was a pain to reverse everything that was done by the return, that's why I suggest to set it to Not installable.

pedrobaeza:
Your case is very specific (same partner in several payments), so no reason for disable it. You can also fix it and propose the patch in a PR.

sandrafig:
Thank you Pedro, I'll take a look at the code and see what I can do (I'm a noob in Python)

pedrobaeza:
Closing due to aging.

PR #84, opened 2016-05-18
8.0 [FIX] [account_payment_return]

sandrafig:
Fixes issue #83 
Changes move_line_ids (many2many) to move_line_id (many2one)


pedrobaeza:
This is not a fix, this is a change in the functionality. We changed it for having the possibility of conciliating several invoices against the same return, so this is a step-back.

:-1:

PR #85, opened 2016-05-23
New module: account_early_payment_discount

cyrilgdn:
I do not know if this is the right place to put this module ?!


account_early_payment_discount/README.rst
```diff
@@ -0,0 +1,63 @@
...
+
+This module adds **Early Payment Discount** informations in Payment Term model.
+
+Look at ``account_operation_rule_early_payment_discount`` module in https://github.com/OCA/bank-statement-reconcile
+for using Early Payment Discount in reconciliation.
+
+Usage
+=====
+
+To configure Payment Term, user must have Adviser access in Accounting & Finance.
```
pedrobaeza, line 17: Put this in Configuration section

account_early_payment_discount/README.rst
```diff
@@ -0,0 +1,63 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=====================================
+Payment Term - Early Payment Discount
+=====================================
+
+This module adds **Early Payment Discount** informations in Payment Term model.
```
pedrobaeza: s/informations/information

ghost:
@cyrilgdn This seams to me like half a feature. You add the fields on payment.term, but you don't add the logic on invoices/payments/account moves. Why so ?

account_early_payment_discount/models/account_payment_term.py
```diff
@@ -0,0 +1,48 @@
...
+
+        'CHECK '
+        '(NOT early_payment_discount OR'
+        ' (NULLIF(epd_nb_days, 0) IS NOT NULL AND'
+        '  NULLIF(epd_discount, 0) IS NOT NULL)'
+        ')',
+
+        _("'Number of day(s)' and 'Discount' fields "
+          "should be filled if 'Early Payment Discount' is checked")
+    )]
```
ghost, line 37: in my opinion, this is overkill. You should instead write this constraint as python code. This would make your code inheritable and easier to maintain.

cyrilgdn:
@dufresnedavid There's another PR in https://github.com/OCA/bank-statement-reconcile/pull/142 which use this payment term information in bank reconciliation.

I preferred to split in these two modules as account_early_payment_discount could be improved in future and the reconciliation part depends on https://github.com/OCA/bank-statement-reconcile/tree/9.0/account_operation_rule .

ghost:
@cyrilgdn I don't get what your point. The reconciliation part ? What other part is there.  The whole use case of payment terms is to reconcile invoice payments. Isn't it ?

Right now, I don't understand your plan, but if you show me the other module, maybe it would make more sence.

cyrilgdn:
@dufresnedavid The other module was linked in my previous message: https://github.com/OCA/bank-statement-reconcile/pull/142

If you prefer, I can try to merge both modules or closed these PRs if you think it is not interesting for OCA.

ghost:
@cyrilgdn I think the feature is very interesting for OCA. I just don't get the need to split it in 2 different modules. It adds a new dependency between 2 oca repos, just in case that the model could be used for other purpose.

I think you should merge the modules into one PR.

PR #87, opened 2016-06-07
[8.0] Add fast install script for account_due_list

ghost:
This pull request intends to speed up tremendously the installation of account_due_list on an existing instance. Otherwise, the installation is very slow when the db contains a lot of account entries. In a production deployment, it could take hours to update the three computed fields on account.move.line.

Give me your feedbacks.

### DIFF
--- a/account_due_list/init_hook.py
+++ b/account_due_list/init_hook.py
@@ -1,0 +1,110 @@
+# -*- coding: utf-8 -*-
+# © 2016 David Dufresne <david.dufresne@savoirfairelinux.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import logging
+
+from openerp import SUPERUSER_ID
+from openerp.api import Environment
+
+
+logger = logging.getLogger(__name__)
+
+
+def pre_init_hook(cr):
+    """
+    The objective of this hook is to speed up the installation
+    of the module on an existing Odoo instance.
+
+    Without this script, if a database has a few hundred thousand
+    journal entries, which is not unlikely, the update will take
+    at least a few hours.
+
+    The pre init script only writes 0 in the field maturity_residual
+    so that it is not computed by the install.
+
+    The post init script sets the value of maturity_residual.
+    """
+    store_field_stored_invoice_id(cr)
+    store_field_day(cr)
+    store_field_maturity_residual(cr)
+
+
+def post_init_hook(cr, pool):
+    env = Environment(cr, SUPERUSER_ID, {})
+    store_field_maturity_residual_post_init(env)
+
+
+def store_field_stored_invoice_id(cr):
+    cr.execute(
+        """
+        ALTER TABLE account_move_line ADD COLUMN stored_invoice_id integer;
+        COMMENT ON COLUMN account_move_line.stored_invoice_id IS 'Invoice';
+        """)
+
+    logger.info('Computing field stored_invoice_id on account.move.line')
+
+    cr.execute(
+        """
+        UPDATE account_move_line
+        SET stored_invoice_id = subquery.invoice_id
+        FROM (
+            SELECT line.id as move_line_id, inv.id as invoice_id
+            FROM account_move_line line
+            LEFT OUTER JOIN account_move move ON line.move_id = move.id
+            LEFT OUTER JOIN account_invoice inv ON move.id = inv.move_id
+        ) AS subquery
+        WHERE account_move_line.id=subquery.move_line_id;
+        """
+    )
+
+
+def store_field_day(cr):
+    cr.execute(
+        """
+        ALTER TABLE account_move_line ADD COLUMN day character varying(16);
+        COMMENT ON COLUMN account_move_line.day IS 'Day';
+        """)
+
+    logger.info('Computing field day on account.move.line')
+
+    cr.execute(
+        """
+        UPDATE account_move_line
+        SET day = subquery.date_maturity
+        FROM (
+            SELECT id, date_maturity FROM account_move_line
+        ) AS subquery
+        WHERE account_move_line.id=subquery.id;
+        """
+    )
+
+
+def store_field_maturity_residual(cr):
+    cr.execute(
+        """
+        ALTER TABLE account_move_line ADD COLUMN maturity_residual
+        double precision;
+        COMMENT ON COLUMN account_move_line.maturity_residual
+        IS 'Residual Amount';
+        """)
+
+    logger.info(
+        'Storing 0 as value to field amount_residual on account.move.line')
+
+    cr.execute(
+        """
+        UPDATE account_move_line SET maturity_residual = 0
+        """)
+
+
+def store_field_maturity_residual_post_init(env):
+    logger.info('Computing field amount_residual on account.move.line')
+
+    for move_line in env['account.move.line'].search([]):
+        sign = (move_line.debit - move_line.credit) < 0 and -1 or 1
+        maturity_residual = move_line.amount_residual * sign
+        env.cr.execute(
+            """
+            UPDATE account_move_line SET maturity_residual = %s WHERE id = %s
+            """, (maturity_residual, move_line.id))
--- a/account_due_list/init_hook.py
+++ b/account_due_list/init_hook.py
@@ -26,6 +26,7 @@ def pre_init_hook(cr):
     The post init script sets the value of maturity_residual.
     """
     store_field_stored_invoice_id(cr)
+    store_field_invoice_user_id(cr)
     store_field_maturity_residual(cr)
 
 
@@ -50,6 +51,26 @@ def store_field_stored_invoice_id(cr):
         FROM account_move AS am, account_invoice AS inv
         WHERE am.id = aml.move_id
         AND am.id = inv.move_id
+        """
+    )
+
+
+def store_field_invoice_user_id(cr):
+    cr.execute(
+        """
+        ALTER TABLE account_move_line ADD COLUMN invoice_user_id integer;
+        COMMENT ON COLUMN account_move_line.invoice_user_id IS
+        'Invoice salesperson';
+        """)
+
+    logger.info('Computing field invoice_user_id on account.move.line')
+
+    cr.execute(
+        """
+        UPDATE account_move_line aml
+        SET invoice_user_id = inv.user_id
+        FROM account_invoice AS inv
+        WHERE aml.stored_invoice_id = inv.move_id
         """
     )
 
--- a/account_due_list/init_hook.py
+++ b/account_due_list/init_hook.py
@@ -46,15 +46,11 @@ def store_field_stored_invoice_id(cr):
 
     cr.execute(
         """
-        UPDATE account_move_line
-        SET stored_invoice_id = subquery.invoice_id
-        FROM (
-            SELECT line.id as move_line_id, inv.id as invoice_id
-            FROM account_move_line line
-            LEFT OUTER JOIN account_move move ON line.move_id = move.id
-            LEFT OUTER JOIN account_invoice inv ON move.id = inv.move_id
-        ) AS subquery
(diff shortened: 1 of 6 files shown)


account_due_list/init_hook.py
```diff
@@ -0,0 +1,121 @@
+# -*- coding: utf-8 -*-
+# © 2016 David Dufresne <david.dufresne@savoirfairelinux.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import logging
+
+from openerp import SUPERUSER_ID
+from openerp.api import Environment
+from openupgradelib import openupgrade
```
pedrobaeza: Don't use openupgradelib for other things except migrations. Anyway, loggued_query is not very useful in this case.

account_due_list/init_hook.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        ALTER TABLE account_move_line ADD COLUMN stored_invoice_id integer;
+        COMMENT ON COLUMN account_move_line.stored_invoice_id IS 'Invoice';
+        """)
+
+    logger.info('Computing field stored_invoice_id on account.move.line')
+
+    cr.execute(
+        """
+        SELECT line.id, inv.id
```
pedrobaeza: You can make this in only one update sentence.

ghost: @pedrobaeza you mean using a subquery ?

pedrobaeza: Try this:

```
UPDATE account_move_line aml
SET stored_invoice_id = inv.id
FROM account_move AS am, account_invoice AS inv
WHERE am.id = aml.move_id
AND am.id = inv.move_id
```

account_due_list/init_hook.py
ghost: Ok, wasn't sure on that point

ghost:
@pedrobaeza I removed the dependency to openupgradelib and simplified the update query.

account_due_list/init_hook.py
```diff
@@ -0,0 +1,110 @@
...
+        """
+        ALTER TABLE account_move_line ADD COLUMN stored_invoice_id integer;
+        COMMENT ON COLUMN account_move_line.stored_invoice_id IS 'Invoice';
+        """)
+
+    logger.info('Computing field stored_invoice_id on account.move.line')
+
+    cr.execute(
+        """
+        UPDATE account_move_line
```
pedrobaeza: My SQL doesn't work?

account_due_list/init_hook.py
```diff
@@ -0,0 +1,110 @@
...
+        COMMENT ON COLUMN account_move_line.maturity_residual
+        IS 'Residual Amount';
+        """)
+
+    logger.info(
+        'Storing 0 as value to field amount_residual on account.move.line')
+
+    cr.execute(
+        """
+        UPDATE account_move_line SET maturity_residual = 0
```
pedrobaeza: This doesn't serve, as the value is not correct. You should put the correct one.

ghost: in fact, I should remove this part, because it is useless. The correct value is set at line 109

account_due_list/init_hook.py
```diff
@@ -0,0 +1,102 @@
...
+        ALTER TABLE account_move_line ADD COLUMN day character varying(16);
+        COMMENT ON COLUMN account_move_line.day IS 'Day';
+        """)
+
+    logger.info('Computing field day on account.move.line')
+
+    cr.execute(
+        """
+        UPDATE account_move_line
+        SET day = subquery.date_maturity
```
pedrobaeza: This doesn't set the day

ghost: yes it does

select day from account_move_line where day is not null limit 10
+------------+
| day        |
|------------|
| 2016-05-05 |
| 2016-07-03 |
| 2016-07-03 |
| 2016-06-02 |
| 2016-07-02 |
| 2016-06-02 |
| 2016-07-01 |
| 2015-09-11 |
| 2015-09-11 |
| 2016-07-01 |
+------------+
SELECT 10
Command Time: 0.000s
Format Time: 0.003s

account_due_list/init_hook.py
ghost: what does not work, what is the error message ?

pedrobaeza: My SQL is:

```
UPDATE account_move_line aml
SET stored_invoice_id = inv.id
FROM account_move AS am, account_invoice AS inv
WHERE am.id = aml.move_id
AND am.id = inv.move_id
```

ghost: @pedrobaeza first, you are missing a join condition. Second, I prefer a subquery. I find it more explicit. I did not try your query. It probably works as well.

pedrobaeza: I'm not omitting anything. A subquery is always less optimal in performance.

ghost: never mind, you're not missing a join condition

account_due_list/init_hook.py
pedrobaeza: Well, I will expect that this is a day, not the full date. And the value is the same, so what's the purpose of this field?

ghost: @pedrobaeza I asked myself the same question. Maybe the module needs a refactoring. I'm only trying to improve the installation, not the module itself.

account_due_list/init_hook.py
ghost: @pedrobaeza Ok, I removed the subqueries. It's about 40% faster.

account_due_list/init_hook.py
pedrobaeza: @eLBati, do you know the reason for this field?

ghost:
@pedrobaeza IMHO the reason for the field day is that the goupby on dates in Odoo groups by month. Maybe some javascript could have done the trick. Maybe this could be changed in another pull request.

pedrobaeza:
@dufresnedavid, yeah, I see. That can be, but there's no need for another field. You can do it with filters in the search view.

angelmoya:
Great work, if you don't planned to do the same init hook for module account_due_list_payment_mode I could do it

angelmoya:
@pedrobaeza ... about day field, this is a stored char field related to a date field, it could not be a problem with users with distinct language, for example
12/01/2016 (english date) = 01/12/2016 (spanish date)

pedrobaeza:
@amoya-dx, I don't understand you.

angelmoya:
And field day is invisible in tree view and is not used in other way.... maybe it is deprecated legacy code

pedrobaeza:
Let's remove it then.

angelmoya:
if it is deleted then it is no longer sifnificant but about languaje and dates, I am not sure but I think that it could be this scenario:

English user create a move line with maturity day January 12, 2016, this is stored on day field in english as "01/12/2016".

Spanish user create a move line with maturity day December 1, 2016, this is stored on day field in spanish as "01/12/2016".

That is because when I get a date field with an user I get the date in this language, or not?

account_due_list/init_hook.py
eLBati: @pedrobaeza @dufresnedavid it should be only used to group by it:
https://github.com/savoirfairelinux/account-payment/blob/139b3de2cdb5b4133187ca748ce617c13db64f3b/account_due_list/views/payment_view.xml#L55

But I think it can be replaced by `date_maturity`. I don't remember why I made it like that.
I only find this commit
http://bazaar.launchpad.net/~domsense/domsense-agilebg-addons/6.1/revision/105
but it is probably a mistake

pedrobaeza: @eLBati, thanks for the clarification. @dufresnedavid, can you then make the switch in view and remove the field?

ghost:
@pedrobaeza @eLBati @amoya-dx Ok, I'm gonna remove the field day in this same PR. Everyone seems to agree that this field is useless.

account_due_list/init_hook.py
ghost: @pedrobaeza will do, juste let me 30 minutes in order to test the result

account_due_list/views/payment_view.xml
```diff
@@ -53,7 +52,6 @@
                 <group expand="0" string="Group By...">
                     <filter string="Partner" icon="terp-partner" domain="[]" context="{'group_by':'partner_id'}"/>
                     <filter string="Invoice" icon="terp-folder-orange" domain="[]" context="{'group_by':'stored_invoice_id'}"/>
-                    <filter string="Due date" icon="terp-go-today" domain="[]" context="{'group_by':'day'}"/>
```
angelmoya, line 56: This filter is userfull, you can replace for this:

```
 <filter string="Due date" icon="terp-go-day" domain="[]" context="{'group_by':'date_maturity:day'}"/>
```

ghost, line 56: @amoya-dx done

ghost:
@pedrobaeza @amoya-dx @eLBati Travis is green. Can we merge now ?

account_due_list/migrations/8.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,18 @@
...
+    ('account_move_line', 'day'),
+}
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    if not version:
+        return
+
+    openupgrade.drop_columns(cr, column_drops)
```
pedrobaeza, line 18: This is not needed, as it's done by ORM itself.

ghost, line 18: @pedrobaeza the orm does not drop columns, only tables.

ghost:
I just realized that i forgot the field invoice_user_id.

account_due_list/migrations/8.0.1.0.0/pre-migration.py
pedrobaeza, line 18: AFAIK it does: https://github.com/odoo/odoo/blob/3ddb446b3bbbaeab17d03900247481a501c7aaf9/openerp/models.py#L2508

pedrobaeza, line 18: Uhm, no, that's for functions. I'm wrong then. Thanks for clarifying.

ghost, line 18: @pedrobaeza this is absolutely another context. Just look a few lines above of the code you pointed. What it does is drop the column when the field changes from stored to not stored.

In my case, the field does not exist anymore.

pedrobaeza, line 18: Yeah, sorry, I mixed things.

JordiBForgeFlow:
:+1: Thanks for the contribution!

JordiBForgeFlow:
I believe that this one is ready to merge.

account_due_list/init_hook.py
```diff
@@ -0,0 +1,97 @@
...
+    logger.info('Computing field invoice_user_id on account.move.line')
+
+    cr.execute(
+        """
+        UPDATE account_move_line aml
+        SET invoice_user_id = inv.user_id
+        FROM account_invoice AS inv
+        WHERE aml.stored_invoice_id = inv.move_id
+        """
+    )
```
omar7r, line 75: inv.id instead of inv.move_id

pedrobaeza, line 75: Right, I'm fixing it directly.

pedrobaeza, line 75: Done in https://github.com/OCA/account-payment/commit/b4ac4d903d72d68ab8f03c9fdd164b6d6356ee7a

omar7r, line 75: Thanks @pedrobaeza

ghost, line 75: @pedrobaeza @omar7r  thanks for the fix

PR #94, opened 2016-07-14
[ADD] account_due_list_days_overdue - refactorization

JordiBForgeFlow:
This is an experiment to try to create dynamic overdue periods...

cc @hbrunn


hbrunn:
see https://github.com/Eficent/account-payment/pull/1

JordiBForgeFlow:
@hbrunn Now it works when you create a new overdue term, but only when you restart the server. Not sure that calling _register_hook on the account.overdue.term in create, write, helps (I tried, with no success).

hbrunn:
try calling https://github.com/OCA/OCB/blob/8.0/openerp/modules/registry.py#L486 instead of the register hook, this signals to all registries they have to reload, probably you're in a multi process environment where not all processes know about your change, and if you're lucky, this also calls the register hook again

JordiBForgeFlow:
@hbrunn Now works like a charm ;) Thanks!!!
cc @elicoidal

account_due_list_days_overdue/README.rst
```diff
@@ -0,0 +1,71 @@
...
+
+Usage
+=====
+
+To use this module, you need to go to:
+
+* Invoicing / Journal Entries / Payments and due list
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
```
elicoidal, line 36: recommended:
:width: 600px

elicoidal:
Good news :)
Will test on runbot when available

JordiBForgeFlow:
@elicoidal @hbrunn Travis is green, with nearly 90% of test coverage :)

account_due_list_days_overdue/README.rst
JordiBForgeFlow, line 36: What do you mean by that?

JordiBForgeFlow:
cc @jgrandguillaume This one is also worth a look.

jgrandguillaume:
Hi,

Thanks for the work, I've just tested it on runbot and works great, even with multiple payment terms and payment.

:+1: 

Regards,

Joël

JordiBForgeFlow:
Thanks @jgrandguillaume. I sweated blood on this one. Thankfully @hbrunn helped me a lot to resolve. A key concept that we now understand is how to dynamically add fields (including computed fields) to a any model.

account_due_list_days_overdue/README.rst
elicoidal, line 36: ignore

elicoidal:
Tested :+1: 

2 small details about the menus:
- configuration entry should be in configuration/miscellaneous after payment terms
- report should be either in customer menu (more accessible) or in reporting/generic/partner (more logical)

elicoidal:
another small nice to have: when group by in the reporting, the total should be displayed (like DR/CR/Residual)

JordiBForgeFlow:
@elicoidal I don't think that it is possible to sum on group by in computed fields. And these fields cannot be stored because the value depends on current date.

I believe that this data model can be used to build a qweb report that shows the total due amounts by periods, grouped by partner.

elicoidal:
That is what I suspected.

eLBati:
Thanks :+1:

eLBati:
@jbeficent what do you think about @elicoidal remarks at https://github.com/OCA/account-payment/pull/94#issuecomment-233118677 ?

JordiBForgeFlow:
IMHO this resolves the need better: https://github.com/OCA/account-payment/pull/94

eLBati:
@jbeficent you are linking this same PR. What did you mean?

PR #95, opened 2016-07-19
[FIX] List view fix

JordiBForgeFlow:
Field 'day' was removed from the due list tree view in a simultaneous PR, so we had to find another field as anchor to introduce the new field in the list view.


pedrobaeza:
:+1:

elicoidal:
:+1:

sbidoul:
:+1:

PR #98, opened 2016-09-04
[IMP] Extends functionality of invoice_number to customers, changing …

Danisan:
…the "supplier_invoice_number" field to "document_number".

I also suggest to change module name to "account_voucher_invoice_number"


pedrobaeza:
Closing as very old with conflicts.

PR #101, opened 2016-10-03
[9.0] account_due_list - port to v9

JordiBForgeFlow:


eLBati:
Thanks @jbeficent , please see travis: it seems the author field is too big

account_due_list/__openerp__.py
```diff
@@ -1,42 +1,30 @@
...
-    'name': "Payments Due list",
-    'version': '8.0.0.1.0',
+    'name': "Receivables and Payables Open Items",
+    'version': '9.0.1.0.0',
     'category': 'Generic Modules/Payment',
-    'author': 'Odoo Community Association (OCA), Agile Business Group, '
-              'Zikzakmedia SL',
+    'author': 'Odoo Community Association (OCA), '
+              'Agile Business Group, '
+              'Serv. Tecnol. Avanzados - Pedro M. Baeza,'
```
pedrobaeza: You can switch this to `Tecnativa`, but please also use a short form for your name: `Eficent`. I think it's enough for your branding and will aid in avoiding these issues.

account_due_list/i18n/de.po
```diff
@@ -1,109 +0,0 @@
-# Translation of OpenERP Server.
```
pedrobaeza, line 1: Why are you removing these translations? For sure any of the terms will match. Let Transifex to make its work and match the same terms and discard in the next synchronization the not valid ones.

account_due_list/migrations/8.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+column_drops = {
+    ('account_move_line', 'day'),
+}
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    if not version:
```
pedrobaeza, line 15: This is what `openupgrade.migrate` does, so you can remove it.

account_due_list/migrations/9.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,16 @@
...
+# © 2015 Eficent Business and IT Consulting Services S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+import logging
+
+logger = logging.getLogger('upgrade')
+
+
+def migrate(cr, version):
+    logger.info("Migrating account_due_list from version %s", version)
```
pedrobaeza, line 11: There's already an integrated message for this

account_due_list/migrations/9.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,16 @@
...
+    if not version:
```
pedrobaeza, line 12: Put the decorator

account_due_list/models/account_move_line.py
```diff
@@ -0,0 +1,49 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openerp import models, fields, api
+
+
+class AccountMoveLine(models.Model):
+    _inherit = 'account.move.line'
+
+    invoice_origin = fields.Char(related='invoice_id.origin',
+                                 string='Source Doc')
```
pedrobaeza, line 19: Non stored fields?

account_due_list/models/account_move_line.py
```diff
@@ -0,0 +1,49 @@
...
+    invoice_date = fields.Date(related='invoice_id.date_invoice',
+                               string='Invoice Date')
+    partner_ref = fields.Char(related='partner_id.ref', string='Partner Ref')
+    payment_term_id = fields.Many2one('account.payment.term',
+                                      related='invoice_id.payment_term_id',
+                                      string='Payment Terms')
+    invoice_user_id = fields.Many2one(
+        comodel_name='res.users', related='invoice_id.user_id',
+        string="Invoice salesperson", store=True)
```
pedrobaeza, line 28: What about the aging field?

JordiBForgeFlow:
Thanks @eLBati . Travis is now green.

cc @aheficent

AaronHForgeFlow (approved):
Tested on runbot :+1:

pedrobaeza (changes requested):

account_due_list/migrations/8.0.1.0.0/pre-migration.py
eLBati: @pedrobaeza @jbeficent shouldn't this file be removed from version 9 of the module?

pedrobaeza: Yeah, you're totally right!

eLBati:
@jbeficent looking at the diff with version 8:
`git diff origin/8.0 Eficent/9.0-eficent-account_due_list account_due_list`
I see translations and fields missing.
Did you follow the procedure described at
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0#howto
?

pedrobaeza:
You shouldn't split the menu in 2 for receivables and payables. You still want to see both together, but with the possibility of filtering. Other thing: have you thought about doing this in a pivot table?

pedrobaeza:
@jbeficent I have found these issues:
- Two separate menus for customer/suppliers. We should keep the unified one with the filters, or at least provide the three choices.
- Not all receivable nor payable accounts are displayed.

pedrobaeza:
@jbeficent, any update on this?

JordiBForgeFlow:
@pedrobaeza Will work on this tomorrow cc @lreficent

JordiBForgeFlow:
Working on this now..

PR #103, opened 2016-10-13
[ADD] account_payment_cc

max3903:


pedrobaeza:
I don't really understand why you make this instead of using bank statements. I see also the module in old API.

bistaray:
@pedrobaeza Can you describe the bank statement option?  The main justification for this module is that when a user buys a Pizza with a Company credit card, they don't owe the Pizza guy, they owe the credit card company.  Users however, want to enter the expense as a Vendor Invoice from the Pizza Guy and attach the receipt to it.  They don't want to enter an Invoice from their Credit Card Vendor until they get one (their bill based on their bank statement).  So this module moves the AP from the Pizza Guy to the Credit Card "Vendor".

pedrobaeza:
My process is:
- Create regular supplier invoice.
- Import my credit card statement with a journal pointing to a credit card account.
- Reconcile each operation with its invoice, moving the balance to the credit card account.
- When the charge is made in my bank account, I import the bank statement and annotate it in the credit card account.
- Any balance different from 0 in the credit card account is an error after the regularization.

bistaray:
Thanks for sharing.

Who is the Vendor in the first step - the Pizza Guy or American Express?

pedrobaeza:
The vendor. For me, American Express is not a payable, but only an account to balance.

andhit-r:
@pedrobaeza your work around is doable with few notes:
1. When supplier invoice reconciled with credit card bank statement the payable won't transfer to credit card company. It should be transfered.
2. Bank statement will not fill **date_maturity** field. It's needed to know when we have to pay credit card.

IMO, journal entry related credit card purchase should be:

## Supplier invoice

Expense _Dr._
    A.P  _Cr._    (with Pizza as partner)

## Credit Card Payment

A.P _Dr._ (with Pizza as Partner)
    Credit Card A.P _Cr._ (with credit card issuer as partner)

## Pay Credit Card Statement

Credit Card A.P _Dr._ (with credit card issuer as Partner)
    Bank _Cr._

That's why i prefer use journal entry to record credit card payment.

@max3903 i'm interested with this module. Why did you close it?

PR #107, opened 2016-10-25
[IMP] SUITE account payment return: Improvements to suite

carlosdauden:
- Add raise "You can not be removed a payment return if state is 'Done'".
- New SEPA return code.
- Add check to match returns after import.
- Spanish translation.

@Tecnativa

### DIFF
--- a/account_payment_return/models/payment_return.py
+++ b/account_payment_return/models/payment_return.py
@@ -108,6 +108,13 @@ class PaymentReturn(models.Model):
         return {'returned_payment': False}
 
     @api.multi
+    def unlink(self):
+        if self.filtered(lambda x: x.state == 'done'):
+            raise UserError(_(
+                "You can not remove a payment return if state is 'Done'"))
+        return super(PaymentReturn, self).unlink()
+
+    @api.multi
     def button_match(self):
         self.mapped('line_ids').filtered(lambda x: (
             (not x.move_line_ids) and x.reference))._find_match()
--- a/account_payment_return_import/i18n/es.po
+++ b/account_payment_return_import/i18n/es.po
@@ -1,31 +1,33 @@
 # Translation of Odoo Server.
 # This file contains the translation of the following modules:
 # * account_payment_return_import
-# 
+#
 # Translators:
 # Sergio Teruel Albert, 2015
 msgid ""
 msgstr ""
 "Project-Id-Version: account-payment (8.0)\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2016-08-15 18:00+0000\n"
-"PO-Revision-Date: 2016-09-16 21:45+0000\n"
-"Last-Translator: OCA Transbot <transbot@odoo-community.org>\n"
-"Language-Team: Spanish (http://www.transifex.com/oca/OCA-account-payment-8-0/language/es/)\n"
+"POT-Creation-Date: 2016-10-25 13:32+0200\n"
+"PO-Revision-Date: 2016-10-25 13:36+0200\n"
+"Last-Translator: Carlos Incaser <carlos@incaser.es>\n"
+"Language-Team: Spanish (http://www.transifex.com/oca/OCA-account-payment-8-0/"
+"language/es/)\n"
+"Language: es\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
-"Content-Transfer-Encoding: \n"
-"Language: es\n"
+"Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
+"X-Generator: Poedit 1.8.7.1\n"
 
 #. module: account_payment_return_import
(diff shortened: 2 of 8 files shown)


account_payment_return/models/payment_return.py
```diff
@@ -108,6 +108,13 @@ def _prepare_invoice_returned_cancel_vals(self):
         return {'returned_payment': False}
 
     @api.multi
+    def unlink(self):
+        if self.filtered(lambda x: x.state == 'done'):
+            raise UserError(_(
+                "You can not be removed a payment return if state is 'Done'"))
```
pedrobaeza: Remove `be`

account_payment_return_import/i18n/es.po
```diff
@@ -83,12 +91,14 @@ msgid "Cancel"
...
 #. module: account_payment_return_import
-#: code:addons/account_payment_return_import/wizard/payment_return_import.py:141
+#: code:addons/account_payment_return_import/wizard/payment_return_import.py:144
 #, python-format
 msgid ""
 "Could not make sense of the given file.\n"
 "Did you install the module to support this type of file?"
-msgstr "No se pudo hacer con sentido de el archivo dado.\n¿Instalaste el modulo para soportar este tipo de archivo?"
+msgstr ""
+"No se pudo hacer con sentido de el archivo dado.\n"
```
pedrobaeza: Trato de usted, no de tú

account_payment_return_import_sepa_pain/data/unpaid_reason_codes.xml
```diff
@@ -42,6 +42,10 @@
         <field name="code">MD01</field>
         <field name="name">No valid Mandate</field>
     </record>
+    <record id="unpaid_reason_MD06" model="payment.return.reason">
+        <field name="code">MD06</field>
+        <field name="name">Disputed authorised transaction</field>
```
pedrobaeza, line 47: This isn't translated?

pedrobaeza (changes requested):

account_payment_return_import/i18n/es.po
carlosdauden: Leo cosas contradictorias al respecto, por eso nunca lo tengo claro...

account_payment_return_import_sepa_pain/data/unpaid_reason_codes.xml
carlosdauden: https://github.com/OCA/account-payment/pull/107/files#diff-e37f5160827a396976dac1201391ad64R61

pedrobaeza: OK

pedrobaeza:
Include a test for unlinking returns for not dropping coverage

carlosdauden:
@pedrobaeza changes done

carlosdauden:
cc @sergio-teruel

pedrobaeza:
Please don't squash while the review process. You only need to squash before merging (when the PR is approved), or it's more difficult to track changes from one review to the next. Reviewer can indicate you if you need to squash or not.

pedrobaeza:
You have to test also a case where there's no error and the payment can be removed.

pedrobaeza (approved):
Thanks for the changes.

To the merger: squash on merging

rafaelbn (approved):
OK! Thanks

sergio-teruel (approved):
Ok..

account_payment_return/models/payment_return.py
```diff
@@ -108,6 +108,13 @@ def _prepare_invoice_returned_cancel_vals(self):
         return {'returned_payment': False}
 
     @api.multi
+    def unlink(self):
+        if self.filtered(lambda x: x.state == 'done'):
+            raise UserError(_(
+                "You can not remove a payment return if state is 'Done'"))
+        return super(PaymentReturn, self).unlink()
```
yajo, line 115: I'd implement this as a [global ir.rule](https://www.odoo.com/documentation/8.0/reference/security.html#record-rules) instead.

yajo (changes requested):

account_payment_return/models/payment_return.py
carlosdauden: It is an option, but usually this is the way used and possibly better performance.

pedrobaeza: The problem is the overlapping of access rules that grants you at the end the unlink. This has to be the way.

pedrobaeza:
Merging

account_payment_return/models/payment_return.py
yajo: Well, I know it's late and after all it works fine this way, but just to clarify, I meant a _global_ rule. Quoting from the above link:

> Global rules are subtractive, they must all be matched for a record to be accessible

pedrobaeza: OK, thanks for clarifying. Odoo doesn't do this kind of things by security rules, so there should be a reason for that. Maybe performance as pointed by Carlos, or maybe to not bypass this check removing access rule.

PR #122, opened 2016-12-28
[9.0] [MIG] account_due_list

JordiBForgeFlow:
Supersedes https://github.com/OCA/account-payment/pull/101


JordiBForgeFlow:
cc @sergio-teruel @pedrobaeza @lreficent @aheficent

AaronHForgeFlow (approved):
Functional review :+1:

LoisRForgeFlow (approved):
Tested on local, LGTM :+1:

pedrobaeza:
Please squash all the "OCA Transbot..." commits together.

JordiBForgeFlow:
@pedrobaeza Can you squash all commits of the same author together, even when there are commits of other authors in the middle? I think that you can squash commits of the same author, as long as they are consecutive.

pedrobaeza:
You can change the order as long as there's no conflict in the diff, which is the usual with "OCA Transbot". I do `git rebase -i origin/9.0` and then move lines in the editor with the `s` option for squashing. We can make a quick hangout to show you.

JordiBForgeFlow:
@pedrobaeza I used `GIT_EDITOR=gedit git rebase -i origin/9.0` and reordered the lines. It worked well. I think that this deserves a wiki page in OCA/maintainer-tools, so that everybody knows how to proceed, do you agree?

pedrobaeza:
Well, that's something of GIT itself, so I'm not sure if we should do that or point to GIT tutorials instead.

JordiBForgeFlow:
@pedrobaeza But the requirement to merge OCA Transbot commits is specific to OCA, and people should be told how to do that.

pedrobaeza:
OK, you're right

JordiBForgeFlow:
I will create the wiki page then. Thanks

pedrobaeza:
About the module, I'm moving my opinion to migrate Odoo's _account_followup_  module and expand it (with payment modes for example) instead of keeping this one now I have tried a bit more the version. What do you think?

JordiBForgeFlow:
IMHO these are two separate topics. account_followup is to be used in communications with the customer or supplier. Provides a summary of what's outstanding. But account_due_list is a tool used by accountants to control internally the receivables and payables. It is very valuable, and I would not remove it.

pedrobaeza:
OK, they are too stretched, but I agree the other module lacks a "general view".

pedrobaeza:
We will migrate the other one for sure.

JordiBForgeFlow:
@pedrobaeza can you merge?

JordiBForgeFlow:
@pedrobaeza What merge option is best in this case? rebase and merge?

pedrobaeza:
No, merge is correct, but let me check that all is correct, because I haven't checked my issues.

JordiBForgeFlow:
The 'merge' feature has disappeared now in Github. It's confusing

account_due_list/migrations/9.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,65 @@
...
+        WHERE aml.stored_invoice_id = inv.id
+        """
+    )
+
+
+def migrate(cr, version):
+    if not version:
+        return
+    store_field_stored_invoice_id(cr)
+    store_field_invoice_user_id(cr)
```
eLBati, line 14: @jbeficent why is this migration script needed? Columns ```invoice_user_id``` and ```stored_invoice_id``` , shouldn't they be already existing if the module is installed on a version 8 DB?

JordiBForgeFlow: This is done to avoid the recomputation during the database migration, which takes an awful lot time.

JordiBForgeFlow:
@pedrobaeza Can we merge this one?

JordiBForgeFlow:
@pedrobaeza we really need this one to be merged.

account_due_list/migrations/8.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
```
pedrobaeza: Remove this file

account_due_list/migrations/9.0.1.0.0/pre-migration.py
pedrobaeza: Don't duplicate code. Just import the functions in pre-init-hook and call them, adjusting the differences in one side or the other.

pedrobaeza (changes requested):
Last sprint with several comments

account_due_list/migrations/9.0.1.0.0/pre-migration.py
JordiBForgeFlow: Ok, just did that

JordiBForgeFlow:
@pedrobaeza I removed the migration from 8 (why?) and centralized the functions that are used by the migration of 9 and init_hook

JordiBForgeFlow:
@pedrobaeza this one is ready to merge

pedrobaeza (approved):

JordiBForgeFlow:
Thanks!!!

PR #128, opened 2017-03-24
[9.0][account_due_list] fix migration script

JordiBForgeFlow:
This PR fixes the migration part of the module.


pedrobaeza:
Sorry, my fault, but you still can import them with relative import. Just put 3 dots instead 2

JordiBForgeFlow:
@pedrobaeza OK, but I will wait until monday. This weekend I have a mig migration and I don't want more heart attaks :)

pedrobaeza (approved):
OK, let's merge as is as this script will be removed in v10.

PR #135, opened 2017-04-14
[FIX] Dependency and documentation

max3903:


account_due_list/__manifest__.py
```diff
@@ -18,11 +18,12 @@
     'author': 'Odoo Community Association (OCA), '
               'Agile Business Group, '
               'Tecnativa,'
-              'Zikzakmedia SL',
+              'Zikzakmedia SL, '
+              'Ursa Information Systems',
```
pedrobaeza: Adding Ursa as author changing only the dependency is not a meaningful co-authoring reason I'm afraid.

lasley (approved):
Agreed with Pedro co-authorship. LGTM otherwise

pedrobaeza (approved):

PR #151, opened 2017-06-20
[ADD] payment_sisow

hbrunn:
Sisow Payment Acquirer
======================

This module was written to support payments via the Dutch payment acquirer [Sisow](https://sisow.nl)
in Odoo.

Configuration
=============

To configure this module, you need to:

- go to Accounting / Configuration / Payments / Payment Acquirers
- find the record with `provider` ``Sisow``
- fill in your `merchant ID` and your `merchant key` as provided by Sisow
- after testing, set the field `Environment` to ``Production`` and click the top right button to make the payment acquirer visible on the website

Known issues / Roadmap
======================

* currently, this module only supports Sisow's iDeal implementation


pedrobaeza:
@hbrunn is this going to be finished?

hbrunn:
it's in production for a customer. But given there doesn't seem much interest in this and I won't fix the tests any time soon, we can just as well close this and reopen whenever we commits more time to this

pedrobaeza:
I'm seeing that the tests that are failing are inside the module. Maybe better to remove them and let this comes to this branch?

hbrunn:
let's not water down OCA's quality standards. Code without tests is not mergable in my book, so I rather close this

pedrobaeza:
OK, I also prefer with tests, but remember that it's not an OCA requirement right now.

hbrunn:
oh, let's clarify this. I read [Tests or scenario are not all green and/or not adapted](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#it-makes-sense-to-be-picky-in-the-following-cases) as 'when coveralls is red, no merge' and that's also what I execute in 'my' repositories which gives me unpleasant discussions every now and then.

I think it's a good thing to demand that new code has at least the coverage as the repo, so I'm fine with the way it is. But if you (and probably the board) disagree with this reading, then I think we should change the wording.

pedrobaeza:
No, I totally agree with you of course. I would try to go to 100% coverage (see https://github.com/OCA/vertical-association that it's totally ours), but again there's the argument about not difficulting newcomers contribution and so on...

PR #152, opened 2017-07-07
[MIG] account_payment_return: Migration to 10.0

chienandalu:
Returned Customers Payment Orders
=================================

This module implements customer receivables returns and allows to send
related reconciled account move lines back to a state where the debt is still
open, and letting history of it.

This module can be extended adding importers that automatically fills the
full returned payment record.

cc @Tecnativa


pedrobaeza (approved):
@luismontalba please review

luismontalba (approved):
Functionality tested on runbot

PR #159, opened 2017-09-20
[FIX] Singleton id issue

max3903:


rmcbean-ursa (approved):
Approved after testing.

smangukiya (approved):
It's OK to proceed!

PR #164, opened 2017-10-30
[ADD] account_payment_term_multi_day: migration scripts

cubells:
If module paydays was installed, we must propagate paydays data to lines.

cc @Tecnativa

### DIFF
--- a/account_payment_term_multi_day/migrations/8.0.1.0.0/post-migration.py
+++ b/account_payment_term_multi_day/migrations/8.0.1.0.0/post-migration.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp.openupgrade import openupgrade
+
+
+def update_paydays(cr):
+    """
+    If column paydays exists in account_payment_term then module paydays was
+    installed. We must propagate paydays to account_payment_term_line to avoid
+    to lose data.
+    """
+    if not openupgrade.column_exists(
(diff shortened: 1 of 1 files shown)


account_payment_term_multi_day/migrations/8.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def update_paydays(cr):
+    """
+    Propagate paydays field to all lines of payment term.
+    """
+    openupgrade.logged_query(
+        cr,
+        """
+        UPDATE account_payment_term_line aptl
```
pedrobaeza: Don't make subqueries:
```
UPDATE account_payment_term_line aptl
SET paydays = apt.paydays
FROM account_payment_term apt
WHERE apt.id = aptl.payment_id
```

account_payment_term_multi_day/migrations/8.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,29 @@
...
+            WHERE apt.id = aptl.payment_id)
+    """)
+
+
+@openupgrade.migrate(no_version=True)
+def migrate(cr, version):
+    if not openupgrade.column_exists(
+            cr, 'account_payment_term', 'paydays'):
+        return
+    # If column paydays exists in account_payment_term then module paydays was
```
pedrobaeza: Make everything in the same method, putting this explanation in the docstring, and letting this main method as only a call to each method (one in this case).

pedrobaeza (changes requested):

pedrobaeza (approved):

pedrobaeza:
Merging optimistically this following OpenUpgrade philosophy.

PR #174, opened 2018-03-09
[10.0] Supplier invoice and Cash discount

benwillig:
Theses modules add the concept of cash discount if an invoice has been paid entirely before a due date.

- ```account_cash_discount_base```: Compute a discount on an invoice based on a rate
- ```account_cash_discount_payment```: Handle the cash discount in a payment order
- ```account_cash_discount_writeoff```: Create a writeoff when confirming the payment order to mark the invoice as 'paid'


benwillig:
cc @chrisdec @sbidoul

account_cash_discount_base/README.rst
```diff
@@ -0,0 +1,95 @@
...
+from the delay. The amount with discount deducted is computed at the same time
+as taxes. This module also adds cash discount informations on customer invoice
+report.
+
+On supplier invoices, you can set the discount percent. For the discount due
+date, you can fill it directly or set a delay. In this case, discount due date
+is still computed at invoice's validation.
+
+You can decide to calculate the discount amount based on the total price taxes
+included or not.
```
sbidoul, line 22: This module was written to allow you to define cash discount (aka early payment discount)
on customer or supplier invoices. 

On customer invoices, you can add a discount delay and
a discount percentage. At invoice validation, the discount due date is generated
from the delay. The amount with discount deducted is computed at the same time
as taxes. This module also adds cash discount informations on customer invoice
report.

On supplier invoices, you can set the discount percentage. For the discount due
date, you can fill it directly or set a delay. In this case, discount due date
is still computed at invoice's validation.

You can decide to calculate the discount amount based on the total price taxes
included or not.

account_cash_discount_base/README.rst
```diff
@@ -0,0 +1,95 @@
...
+
+
+Installation
+============
+
+To install this module, you need to:
+
+ * Click on install button
```
sbidoul: Remove this section

account_cash_discount_base/models/account_invoice.py
```diff
@@ -0,0 +1,119 @@
...
+    )
+    def _compute_discount_amount(self):
+        for rec in self:
+            discount_amount = 0.0
+            if rec.discount_percent != 0.0:
+                base_amount_type = \
+                    rec.company_id.cash_discount_base_amount_type
+                base_amount = (
+                    base_amount_type == 'untaxed' and
+                    rec.amount_untaxed or rec.amount_total)
```
sbidoul, line 86: a good old if then else is easier to read

account_cash_discount_base/models/account_invoice.py
```diff
@@ -0,0 +1,119 @@
...
+                discount_amount = (
+                    base_amount * (0.0 + rec.discount_percent / 100))
```
sbidoul: why 0.0 + ... ?

account_cash_discount_base/README.rst
```diff
@@ -0,0 +1,95 @@
...
+   :target: https://runbot.odoo-community.org/runbot/96/10.0
+
+For further information, please visit:
+
+ * https://www.odoo.com/forum/help-1
+
+Known issues / Roadmap
+======================
+
+* If you change the company base amount type, discount amount will change on old invoices as it's a simple compute method.
```
sbidoul: This sounds dangerous. I think it's better to freeze these amounts when validating the invoice. As the readme says: `The amount with discount deducted is computed at the same time as taxes`.

account_cash_discount_base/README.rst
```diff
@@ -0,0 +1,95 @@
...
+To configure the base amount type, you need to:
+
+#. Go to your companies
+#. Set the cash discount base amount type (including taxes or not)
+
+You can also configure default discount percent and delay on the payment term.
+
+.. figure:: path/to/local/image.png
+   :alt: alternative description
+   :width: 600 px
```
sbidoul: Remove or put a real screenshot.

account_cash_discount_base/models/account_invoice.py
```diff
@@ -0,0 +1,119 @@
...
+
+    @api.multi
+    @api.onchange(
+        'discount_delay',
+    )
+    def _onchange_discount_delay(self):
+        date_today = datetime.today()
+        for rec in self:
+            if rec.discount_amount == 0 or rec.discount_delay == 0:
+                continue
```
sbidoul, line 188: Why is this test necessary?

account_cash_discount_base/models/account_invoice.py
```diff
@@ -0,0 +1,119 @@
...
+    @api.multi
+    def action_move_create(self):
+        res = super(AccountInvoice, self).action_move_create()
+        for inv in self:
+            inv._onchange_discount_delay()
+            if not inv.discount_due_date and inv.discount_amount != 0.0:
+                raise UserError(_(
+                    "You can't set a discount amount if there is no "
+                    "discount due date. (Invoice ID: %s)"
+                ) % (inv.id,))
```
sbidoul, line 227: inv.id is not really helpful for the user (use inv.number instead, at this one it must have one?)

account_cash_discount_payment/README.rst
```diff
@@ -0,0 +1,67 @@
...
+
+This module also offers to search by discount due date to add invoices on
+payment order and add discount amount on payment line.
+
+Installation
+============
+
+To install this module, you need to:
+
+ * Click on install button
```
sbidoul: remove this section

account_cash_discount_payment/README.rst
```diff
@@ -0,0 +1,67 @@
...
+=============================
+
+This module was written to extend Account Cash Discount Base module to allow
+you to add some invoices on payment order considering the cash discount on it.
+Then if you add an invoice on a payment order and if there is a cash discount
+on it. The amount on payment line will be the residual amount discount deducted,
+provided that the due date has not expired.
+
+This module also offers to search by discount due date to add invoices on
+payment order and add discount amount on payment line.
```
sbidoul, line 15: Extend Account Cash Discount Base and Payment Order modules to let you
add invoices on payment order considering the cash discount on it.
When an invoice with cash discount is added to a payment order,
he amount proposed on payment line will be the residual amount discount deducted,
provided that the due date has not expired.

This module also offers to search by discount due date to add invoices on
payment order and add discount amount on payment line.

account_cash_discount_payment/models/account_payment_line.py
```diff
@@ -0,0 +1,45 @@
...
+    )
+    discount_due_date = fields.Date(
+        compute='_compute_discount_due_date',
+    )
+    discount_amount = fields.Monetary(
+        default=0.0
+    )
+    original_amount_currency = fields.Monetary(
+        readonly=True,
+    )
```
sbidoul, line 30: Can't we do the economy of this field? I think we can rely on the invoice total and enable the cash discount payment only if the invoice residual is equal to the invoice total. Ie if there is no invoice, disable pay with discount, and if the discount is partially paid, disable pay with discount. @chrisdec

account_cash_discount_payment/wizards/account_payment_line_create.py
```diff
@@ -0,0 +1,34 @@
...
+
+    _inherit = 'account.payment.line.create'
+
+    date_type = fields.Selection(
+        selection_add=[
+            ('discount_due_date', _("Discount Due Date")),
+        ],
+    )
+    cash_discount_date_start = fields.Date()
+    cash_discount_date_end = fields.Date()
```
sbidoul: Do we need start + end? For due date, the test is `('date_maturity', '<=', self.due_date)`, I think we can do the same for discount due date, it's easier and it's the most common use case.

account_cash_discount_write_off/README.rst
```diff
@@ -0,0 +1,87 @@
...
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===============================
+Account Cash Discount Write Off
+===============================
+
+Create an automatic write-off for payment with discount on the payment order
+confirmation. If the cash discount amount is computed based on the total
+amount, the created write-off will also contains tax adjustments. This
```
sbidoul, line 11: Better have an independent flag to control tax adjustments.

account_cash_discount_write_off/models/account_payment_order.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @api.multi
+    def generated2uploaded(self):
+        res = super(AccountPaymentOrder, self).generated2uploaded()
+        for order in self:
+            order.create_cash_discount_write_off()
+        return res
+
+    @api.multi
+    def create_cash_discount_write_off(self):
```
sbidoul: prefix these methods with `_`

account_cash_discount_write_off/models/account_payment_order.py
```diff
@@ -0,0 +1,44 @@
...
+            line.account_id == payment_line.move_line_id.account_id
+        )
+        lines_to_reconcile.reconcile()
+
+    @api.multi
+    def create_payment_line_discount_write_off_move(self, payment_line):
+        self.ensure_one()
+        move_values = payment_line.get_cash_discount_writeoff_move_values()
+        move = self.env['account.move'].create(move_values)
+        move.post()
```
sbidoul: Post depending on payment mode config: https://github.com/OCA/bank-payment/blob/e752ebe943eb410f97fd419945b5a5fe45b0568a/account_payment_order/models/account_payment_order.py#L497-L498.

account_cash_discount_write_off/models/account_payment_line.py
```diff
@@ -0,0 +1,89 @@
...
+        self.ensure_one()
+        move_line = self.move_line_id
+        has_partial_reconciliation = (
+            bool(move_line.matched_debit_ids) or
+            bool(move_line.matched_credit_ids))
+        return (
+            self.pay_with_discount and
+            has_partial_reconciliation and
+            move_line.invoice_id
+        )
```
sbidoul, line 40: `not move_line.reconciled` instead of `has_partial_reconciliation`?

account_cash_discount_write_off/models/account_payment_line.py
```diff
@@ -0,0 +1,89 @@
...
+from odoo import api, models, _
+from odoo.exceptions import UserError
+
+
+class PaymentLine(models.Model):
+
+    _inherit = 'account.payment.line'
+
+    @api.multi
+    def check_cash_discount_write_off_creation(self):
```
sbidoul: `_` prefix

account_cash_discount_write_off/models/account_payment_line.py
```diff
@@ -0,0 +1,89 @@
...
+            has_partial_reconciliation and
+            move_line.invoice_id
+        )
+
+    @api.multi
+    def get_cash_discount_writeoff_move_values(self):
+        self.ensure_one()
+        move_line = self.move_line_id
+        partner = move_line.partner_id
+        partner = partner._find_accounting_partner(partner)
```
sbidoul: not necessary: the partner on a move line is the accounting partner

account_cash_discount_write_off/models/account_payment_line.py
```diff
@@ -0,0 +1,89 @@
...
+        invoice = move_line.invoice_id
+        company = invoice.company_id
+        tax_adjustment = company.cash_discount_base_amount_type == 'total'
```
sbidoul: use an independent flag for that

account_cash_discount_write_off/models/account_payment_line.py
```diff
@@ -0,0 +1,89 @@
...
+
+        woff_account = False
+        woff_journal = False
+
+        if invoice:
```
sbidoul, line 60: why this test?

account_cash_discount_write_off/models/account_payment_line.py
sbidoul: perhaps we should do this writeoff only when the amound paid is exactly the cash discount amount on the invoice?

account_cash_discount_write_off/models/account_payment_line.py
```diff
@@ -0,0 +1,89 @@
...
+            raise UserError(
+                _("You have to fill in journal and account for cash discount "
+                  "write-off on the company.")
+            )
+
+        move_line_name = _("Cash Discount Write-Off")
+        supplier_account_amount = invoice.discount_amount
+        discount_amount_credit = supplier_account_amount
+
+        lines_values = list()
```
sbidoul: s/list()/[]/

account_cash_discount_write_off/models/account_payment_line.py
```diff
@@ -0,0 +1,89 @@
...
+        lines_values.append({
+            'partner_id': partner.id,
+            'name': move_line_name,
+            'debit': supplier_account_amount,
+            'account_id': move_line.account_id.id,
+        })
+
+        if tax_adjustment:
+            for tax_line in invoice.tax_line_ids:
```
sbidoul: Perhaps work on the move lines that have a tax_id field instead of relying on the invoice tax line?

sbidoul (commented):
Looks very clean :) A few comments and questions.

account_cash_discount_write_off/README.rst
benwillig: But would it be coherent to compute the discount based on the price (excluding taxes) and having the "Tax Adjustments" checked. What should we do in theses situations ?

account_cash_discount_base/models/account_invoice.py
benwillig: I should check the ```discount_percent``` instead of the ```discount_amount```. This test was to prevent to compute a discount due date if it was not needed.

But I'm wondering if we should add a boolean field which would tells if we pay the invoice using the discount or not. If this boolean equals to True, we could maybe force the user to fill the discount_percent and discount_delay.

account_cash_discount_base/models/account_invoice.py
benwillig: We can't use the number of the invoice because when the message is displayed, we rollback the transaction so we lose the invoice number.

account_cash_discount_payment/wizards/account_payment_line_create.py
benwillig: I don't know if start/end range is really needed, we should ask @chrisdec. But in my opinion, we could have a single date indeed.

account_cash_discount_write_off/models/account_payment_line.py
benwillig: I agree that this method is called only when ```check_cash_discount_write_off_creation``` returns true. And this method returns True if the move line is linked to an invoice. So yes, this test might have no sense. but then I'm sure there won't be a huge stacktrace in the future because of a specific case.

account_cash_discount_write_off/models/account_payment_line.py
benwillig: It would be more complicated I think. Because, based on the ```tax_ids``` field, we should get the corresponding move lines, and we could finally compute the amount.

account_cash_discount_base/models/account_invoice.py
benwillig: :alien: I think I was in an other dimension ^^

account_cash_discount_base/README.rst
benwillig: Should we store the ```cash_discount_base_amount_type``` field on the invoice and populate it with an onchange? Or just set the discount fields as stored and remove the method dependency on ```company_id.cash_discount_base_amount_type```?

benwillig:
I uptated my code based on your reviews and the discussions we had. To give a little summary of what I did:
- Invoice discount amounts are now stored
- The pay_with_discount flag on the payment line can only equals to True when the payment line is linked to an invoice which has a discount
- When changing the pay_with_discount flag, the amount is changed based on the invoice total and invoice discount
- A boolean flag has been added to know if we do tax adjustments. At the moment, it's a computed method which will simply assing True if the base amount type for the discount include all taxes. If it's change in the future, there will be less work to do
- The tax adjustments are now created based on the tax move lines
- Prefix some methods with a _
- Updated readmes

sbidoul:
Runbot is orange due to a warning on an onchange.

chrisdec:
Here are the results of first Functional tests:

1.	Could you please add a button on payment order lines to check or uncheck “Pay with discount”. It’s quite annoying to open the line each time you want to change the checkbox. 
2.	When you add journal items to payment orders, if day date < discount date,  the checkbox “Pay with discount” is not checked (Good) but the amount on payment order line is the amount with discount (Not good). 
3.	When you select move lines to put in your payment order, you don’t see cash discount amount. It could be great to add cash discount amount and discount due date in this screen (like in V8). 
4.	When you change “Bill date” on invoice, discount due date must be (re-)computed. 
5.	I can’t find a way to change VAT Base. The V8 module “account_invoice_manual_tax" is not available in V10?
6.	When you work with “Including all taxes”, you have to adapt base taxes and not only taxes on writeoff journal entry.

Thanks :)

benwillig:
4. At the moment, we only use the invoice date to compute the discount due date. Should it be based only on the bill date?
5. No at the moment, this module is not going to be migrated.
6. So if we need to adapt base taxes, how would the invoice journal entry be?

account_cash_discount_write_off/models/account_payment_line.py
```diff
@@ -0,0 +1,88 @@
...
+                    'credit': tax_move_line.debit > 0 and amount or 0.0,
+                    'account_id': tax_move_line.account_id.id,
+                    'tax_line_id': tax_move_line.tax_line_id.id,
+                })
+
+        lines_values.append({
+            'partner_id': partner.id,
+            'name': move_line_name,
+            'credit': discount_amount_credit,
+            'account_id': woff_account.id,
```
sbidoul: @chrisdec @benwillig so here we should add the base taxes (tax_ids) in the writeoff? And therefore split the writeoff in several pieces for each base tax used?

chrisdec: Yes. We have to split the writeoff in several lines but not in several pieces I think.

benwillig:
@chrisdec I updated the module based on your review.

chrisdec:
Hello, 

Here are the results of second Functional tests:

1. Why the button on payment order lines is "+" and "-" and not a checkbox? I think a checkbox would be simpler if it's possible

2. Point 3 of old tests is not good for me : Now, I see the discount amount and discount due date in the "Selected move lines to create Transactions" table (which is good :)) but I would like to see those fields in the "Add : move lines window" if it's possible. 

3. When you work with "Including tax" option, you don't use the write-off account defined on company (debit and credit = zero) but when you work with "Excluding tax" option, you use it. For me, it's not good. The write-off amount must be used in both option. Write-off account must replaced the expense account in the write-off journal entry. 

4. When you select move lines to add to payment order, if you choose to use "Discount due date", normally, the proposed move lines are all move lines with discount due date > selected date. Here, I also see move lines with discount due date < selected date when I click on "Add an item"

Thanks :)

benwillig:
@chrisdec Everything should be fixed now. Could you check again?

chrisdec:
Hello, 

Functional tests are ok for me. 

Thanks :)

rafaelbn:
Interesting @carlosdauden @pedrobaeza

account_cash_discount_base/README.rst
```diff
@@ -0,0 +1,80 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.png
+    :alt: License: AGPL-3
+
+==========================
+Account Cash Discount Base
+==========================
+
+This module was written to allow you to define cash discount (aka early payment discount)
+on customer or supplier invoices.
```
rafaelbn, line 9: @benwillig @sbidoul is this similar to https://github.com/OCA/account-payment/pull/184 ?

sbidoul: Hi @rafaelbn ! This is the same domain, for sure. #184 seems to be about configuring the payment term only. Here we make it easy to pay supplier invoices with early payment discount and account_payment_order (including automatic reconciliation, proper handling of taxes with configurable rules etc). So there is little overlap it seems. At some point we considered using payment terms, which we rejected (I don't remember all the details now but one of the reason is that it was much more complex to handle)?

account_cash_discount_base/models/account_invoice.py
```diff
@@ -0,0 +1,145 @@
...
+READONLY_STATES = {
+    'draft': [('readonly', False)],
+}
+
+
+class AccountInvoice(models.Model):
+
+    _inherit = 'account.invoice'
+
+    discount_percent = fields.Float(
```
rousseldenis, line 24: @benwillig I would have added
```python
digits=dp.get_precision('Discount')
```

account_cash_discount_base/models/account_payment_term.py
```diff
@@ -0,0 +1,17 @@
...
+
+from odoo import fields, models
+
+
+class AccountPaymentTerm(models.Model):
+
+    _inherit = 'account.payment.term'
+
+    discount_percent = fields.Float(
+        string='Discount (%)'
```
rousseldenis: @benwillig Same as above about dp

rousseldenis (changes requested):
Code review

rousseldenis (approved):

account_cash_discount_base/models/account_invoice.py
rousseldenis: Solved

rousseldenis:
@sbidoul @rafaelbn Could you update your review as we could merge this?

rafaelbn (approved):

pedrobaeza:
Can you please squash commit history a bit?

benwillig:
@pedrobaeza Still 18 commits but it's better than 54 I guess ^^

pedrobaeza (dismissed):
What I don't get is the need of the extra module ` account_cash_discount_write_off` instead of an option at payment mode level.

benwillig:
@pedrobaeza This could be an option indeed. But the code is more clear and it will be easier for the migration.

rousseldenis:
I think we can move forward on this as it was proposed quite a year ago. Workin on several production environments.

pedrobaeza:
@rousseldenis I agree to merge base module and the payment one, but not the third, so we can:

- Split these first 2 here, and merge, leaving the third.
- Merge already the 3rd one into the second in this PR, and merge all.

rousseldenis:
@pedrobaeza IMHO there is no added value doing this. Things are well functionnally splitted.

pedrobaeza:
The third module has no reason for being split, as there are no different dependencies and it's as simple as having the field in the parent one, and leaving empty means that there's no writeoff. Over-engineering things with lots of modules has not too much sense IMO.

sbidoul:
@pedrobaeza IMO you are going too far in blocking this. 

The modularity does not harm at all. Ok, you prefer a slightly different design, but the current design has nothing wrong in itself. I would have been fine if you had brought this topic when we were creating the module, but now it's very late. 

The change you propose can be done later in another PR. There is good test coverage, it will be easy to refactor.

If we want to grow the community, we must remove roadblocks. And you must know these kind of blocking comments discourage even my colleagues at ACSONE from contributing to OCA. So imagine the effect it must have on less motivated or new contributors.

So please let's merge this: it's working well and has proven useful in at least 3 different countries.
I'm happy to leave it in Beta status until the redesign you propose has been addressed.

Finally, note we could perfectly have merged this before because we had enough approvals. We are just refraining from merging modules with only approvals from our own company.

pedrobaeza:
If we merge, later then the argument would be that this can't be removed in a stable version. That's why I block it. Any way, I dismiss my review and do whatever you think better, but not with my approval.

sbidoul:
Thanks @pedrobaeza 

We have two approvals from functional people from different companies, and two technical approvals from Acsone, and a remark on modularity by Pedro.

So let's merge in beta status.

PR #176, opened 2018-03-23
[11.0][MIG] account_payment_return

luismontalba:
Returned Customers Payment Orders
==============================
This module implements customer receivables returns and allows to send
related reconciled account move lines back to a state where the debt is still
open, and letting history of it.

Usage
--------
* Go to Invoicing > Sales > Customer Payment Returns, and create a new
record, register on each line a paid (reconciled) receivable journal item,
and input the amount that is going to be returned.
* Another option to fill info is setting references and click match button to
find matches with invoices, move lines or moves. This functionality is extended
by other modules as *account_payment_return_import_sepa_pain*
* Next, press button "Confirm" to create a new move line that removes the
balance from the bank journal and reconcile items together to show payment
history through it.
* After confirmation you can access from the payment form view to the move
created.

cc @Tecnativa


account_payment_return/README.rst
```diff
@@ -0,0 +1,75 @@
...
+related reconciled account move lines back to a state where the debt is still
+open, and letting history of it.
+
+This module can be extended adding importers that automatically fills the
+full returned payment record.
+
+Usage
+=====
+
+Go to Invoicing > Sales > Customer Payment Returns, and create a new
```
elicoidal: Please create enumerated list

account_payment_return/README.rst
```diff
@@ -0,0 +1,75 @@
...
+record, register on each line a paid (reconciled) receivable journal item,
+and input the amount that is going to be returned.
+
+Another option to fill info is setting references and click match button to
```
elicoidal: same here: it increases visibility

account_payment_return/models/__init__.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
```
elicoidal: remove coding in all files (python 3)
Remove copyright info in all init.py files

account_payment_return/models/account_journal.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class AccountJournal(models.Model):
+    _inherit = "account.journal"
+
+    default_expense_account_id = fields.Many2one(
+        comodel_name='account.account', string='Default Expense Account',
```
elicoidal: string is not consistent with help.
I suggest "Default Commission Account"

account_payment_return/models/account_journal.py
```diff
@@ -0,0 +1,17 @@
...
+        help='Default account for commission expenses')
+    default_expense_partner_id = fields.Many2one(
+        comodel_name="res.partner", string="Default Expense Partner",
```
elicoidal: same here

account_payment_return/models/payment_return.py
```diff
@@ -0,0 +1,372 @@
...
+            duplicate_lines = self.env['payment.return.line'].search([
+                ('move_line_ids', 'in', all_move_lines.ids),
+                ('return_id.state', '=', 'done'),
+            ])
+            if duplicate_lines:
+                for line in duplicate_lines:
+                    append_error(line)
+        if error_list:
+            raise UserError(
+                _("Payment reference must be unique!"
```
elicoidal: remove '!'

elicoidal (approved):
some details.
LGTM no test

account_check_printing_report_base/tests/test_check_printing_report.py
```diff
@@ -135,4 +135,4 @@ def test_check_printing_with_layout(self):
             payment.print_checks()
         except UserError as e:
             pass
-        self.assertEquals(e, False)
+        self.assertEqual(e, False)
```
pedrobaeza: Why changing this?

luismontalba: Well, it's changed when you run ```2to3 -wnj4 --no-diffs . ``` as suggested here:
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

pedrobaeza: Yeah, but you must run it not on the entire repository, but on your module folder.

pedrobaeza: seleccion_001

luismontalba: You are right, it was a mistake, I'll fix it

account_payment_return/README.rst
```diff
@@ -16,22 +16,22 @@ full returned payment record.
 Usage
 =====
 
-Go to Invoicing > Sales > Customer Payment Returns, and create a new
-record, register on each line a paid (reconciled) receivable journal item,
-and input the amount that is going to be returned.
+1. Go to Invoicing > Sales > Customer Payment Returns, and create a new
```
pedrobaeza: You don't need to put manually numbers. Putting `#. `, numbers are auto-assigned on rendering.

account_payment_return/models/account_journal.py
```diff
@@ -9,9 +8,9 @@ class AccountJournal(models.Model):
     _inherit = "account.journal"
 
     default_expense_account_id = fields.Many2one(
-        comodel_name='account.account', string='Default Expense Account',
+        comodel_name='account.account', string='Default Commission Account',
```
pedrobaeza: Why changing this? Commission is more for sales, not for the meaning we are thinking (comisión).

account_payment_return/models/account_journal.py
pedrobaeza: Commission is more for sales terminology. I don't think it's correct for an expense.

account_payment_return/models/account_journal.py
luismontalba: I think both terms, bank expenses an bank commissions are right, I changed it because @elicoidal suggested it and I suppose he has better feeling than me about the best term in English language.

account_payment_return/README.rst
luismontalba: Changes done

account_payment_return/models/account_journal.py
elicoidal: @pedrobaeza @luismontalba I suggested `commission` instead of `expense` to avoid the confusion with standard Odoo `expense` account.
Alternative would `Charges` (this is what appears in my bank statement) or `Fees`

pedrobaeza: I like Charges!

luismontalba: OK, I'll put Charges, It's a good option

luismontalba:
Changes done

account_payment_return/models/account_journal.py
```diff
@@ -8,9 +8,9 @@ class AccountJournal(models.Model):
     _inherit = "account.journal"
 
     default_expense_account_id = fields.Many2one(
-        comodel_name='account.account', string='Default Commission Account',
-        help='Default account for commission expenses')
+        comodel_name='account.account', string='Default Charge Account',
```
pedrobaeza: It should be in plural.

account_payment_return/i18n/es.po
```diff
@@ -96,23 +111,23 @@ msgstr "Devoluciones de cobros"
...
-msgstr "Cuenta por defecto para comisiones"
+msgid "Default Charge Account"
+msgstr "Cuenta del cargo por defecto"
 
 #. module: account_payment_return
 #: model:ir.model.fields,field_description:account_payment_return.field_account_journal_default_expense_partner_id
-msgid "Default Commission Partner"
-msgstr "Empresa por defecto para comisiones"
+msgid "Default Charge Partner"
+msgstr "Empresa del cargo por defecto"
```
pedrobaeza: Don't change the previous translation, although the source term changes.

pedrobaeza (changes requested):
Please also change the source term in all current translations

luismontalba:
Changes done

pedrobaeza (approved):

PR #184, opened 2018-04-13
[11.0][MIG] account_early_payment_discount

asaunier:
Migration of a module proposed in 9.0, see https://github.com/OCA/account-payment/pull/85


asaunier:
I don't understand why Travis complains:
```
2018-04-13 15:26:47,957 5038 INFO openerp_test odoo.addons.account_early_payment_discount.tests.test_payment_term: test_constraint__failed (odoo.addons.account_early_payment_discount.tests.test_payment_term.TestPaymentTerm)
2018-04-13 15:26:47,979 5038 ERROR openerp_test odoo.sql_db: bad query: b'INSERT INTO "account_payment_term" ("id", "early_payment_discount", "name", "company_id", "sequence", "active", "create_uid", "write_uid", "create_date", "write_date") VALUES(nextval(\'account_payment_term_id_seq\'), true, \'Unittest payment term\', 1, 10, true, 1, 1, (now() at time zone \'UTC\'), (now() at time zone \'UTC\')) RETURNING id'
ERROR: new row for relation "account_payment_term" violates check constraint "account_payment_term_early_payment_discount"
DETAIL:  Failing row contains (6, Unittest payment term, t, null, 1, 10, 1, 2018-04-13 15:26:47.958971, 1, 2018-04-13 15:26:47.958971, t, null, null, null).
```
Test ``test_constraint__failed`` indeed performs an illegal action by creating a payment term with no ``epd_nb_days`` and ``epd_discount``, breaking the [early_payment_discount constraint](https://github.com/asaunier/account-payment/blob/26357314b607ac512719bec2dd8ad856a60ad53e/account_early_payment_discount/models/account_payment_term.py#L29-L32).

But that's the goal of the test to make sure an integrity error is raised in such a case and a special ``assertRaises`` assertion is actually used:
https://github.com/asaunier/account-payment/blob/26357314b607ac512719bec2dd8ad856a60ad53e/account_early_payment_discount/tests/test_payment_term.py#L12

asaunier:
8e1726d fixes the problem reported above when running the tests.

It seems the ``maintainer-quality-tools`` think tests have failed when an ``ERROR`` keyword is found in the logs. In our case the ``ERROR`` is an expected integrity error when testing an illegal action that breaks a constraint (the test actually passes).

Before this change in Odoo 9 months ago https://github.com/odoo/odoo/commit/c4e9b854ff22af31bbbd88d1414e77d27eaed7ea (see odoo/sql_db.py) such logs were considered as ``INFO``, not ``ERROR``.

8e1726d then prevents the error from being logged using cursor's ``_default_log_exceptions`` property, introduced by https://github.com/odoo/odoo/commit/e85f246469d6dbf5c9f801e20babd7d13b2c05f9

I have added an issue in ``maintainer-quality-tools`` repo at https://github.com/OCA/maintainer-quality-tools/issues/542

This PR is ready for review!

account_early_payment_discount/tests/test_payment_term.py
```diff
@@ -0,0 +1,47 @@
...
+
+class TestPaymentTerm(TransactionCase):
+
+    def test_constraint__failed(self):
+        # The following setting makes sure the integrity error below will not
+        # appear in the logs, which would make the maintainer-quality-tools
+        # report failed tests.
+        # For more details see:
+        # https://github.com/odoo/odoo/commit/e85f246469d6dbf5c9f801e20babd7d13b2c05f9
+        self.env.cr._default_log_exceptions = False
```
simahawk: remove this and decorate the method w/ `@mute_logger('odoo.sql_db')`

account_early_payment_discount/README.rst
```diff
@@ -0,0 +1,63 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
simahawk: use png

simahawk (changes requested):

tschanzt (approved):

asaunier:
@simahawk @tschanzt Thanks for the feedback and tips!
I have updated/rebased the PR to use ``@mute_logger('odoo.sql_db')`` instead and the PNG file for the AGPL icon.

yvaucher (approved):

feketemihai:
@asaunier Did you check the PR #174 , it basically adds the same concept, but in more complex, handle the early payment writeoff, with the module that you propose you only have the info on the payment term, so we will still need a module for writeoff the discount. I suggest to port that one since is handling the same requirement but in a more complex workflow.

simahawk (approved):

account_early_payment_discount/models/account_payment_term.py
```diff
@@ -0,0 +1,47 @@
...
+        'early_payment_discount',
+
+        'CHECK '
+        '(NOT early_payment_discount OR'
+        ' (NULLIF(epd_nb_days, 0) IS NOT NULL AND'
+        '  NULLIF(epd_discount, 0) IS NOT NULL)'
+        ')',
+
+        _("'Number of day(s)' and 'Discount' fields "
+          "should be filled if 'Early Payment Discount' is checked")
```
grindtildeath: s/should/must as it's an imperative.

account_early_payment_discount/models/account_payment_term.py
```diff
@@ -0,0 +1,47 @@
...
+    )]
+
+    @api.depends()
```
grindtildeath: It seems you can remove this one.

account_early_payment_discount/tests/test_payment_term.py
```diff
@@ -0,0 +1,44 @@
...
+                'early_payment_discount': True,
+            })
+
+    def test_constraint__good(self):
+        self.env['account.payment.term'].create({
+            'name': 'Unittest payment term',
+            'early_payment_discount': True,
+            'epd_nb_days': 5,
+            'epd_discount': 2
+        })
```
grindtildeath, line 28: asserting something would be nice.

account_early_payment_discount/README.rst
```diff
@@ -0,0 +1,63 @@
...
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=====================================
+Payment Term - Early Payment Discount
+=====================================
+
+This module adds **Early Payment Discount** information in Payment Term model.
+
+Look at ``account_operation_rule_early_payment_discount`` module in https://github.com/OCA/bank-statement-reconcile
```
grindtildeath: The module will be called 'account_reconcile_rule_early_payment_discount' (to be confirmed by @fclementic2c ) but its repo is now OCA/account-reconcile.

account_early_payment_discount/views/account_payment_term.xml
```diff
@@ -0,0 +1,50 @@
...
+                    </div>
+                </group>
+            </field>
+        </field>
+    </record>
+
+    <!-- Tree view -->
+    <record model="ir.ui.view" id="view_payment_term_tree">
+        <field name="name">account.payment.term.tree</field>
+        <field name="model">account.payment.term</field>
```
grindtildeath: Why don't you inherit `account.view_payment_term_tree` ?

account_early_payment_discount/views/account_payment_term.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <!-- Form view -->
+    <record id="view_payment_term_form" model="ir.ui.view">
+        <field name="model">account.payment.term</field>
```
grindtildeath: A name is missing for this view

grindtildeath (changes requested):
Going to try it out migrating account_operation_rule_early_payment_discount ;)

account_early_payment_discount/tests/test_payment_term.py
asaunier: This test simply checks that it is possible to create a payment term if the mandatory attributes 'epd_nb_days' and 'epd_discount' are provided. Which is not the case in the previous test where those attributes are omitted, causing an integrity error.
The assert is then implicit here.

account_early_payment_discount/views/account_payment_term.xml
asaunier: I guess you mean https://github.com/odoo/odoo/blob/dccef8e41d45dd5150e4112aa3a2b19ba698dc62/addons/account/views/account_view.xml#L1631-L1640

I don't have the answer to your question: I have only migrated the module, not authored it :)
Then I don't know if there is a reason for replacing the view here.

account_early_payment_discount/views/account_payment_term.xml
asaunier: Is it really required?

asaunier:
@grindtildeath Thanks for reviewing!
I have just pushed a fixup taking into account some of your suggestions.

account_early_payment_discount/tests/test_payment_term.py
grindtildeath: IMO you should then assert the field on the created record has the value you provided. If someone breaks the creation doing any change the test will still pass. 
Moreover "Explicit is better than implicit" PEP 20 :wink:

account_early_payment_discount/views/account_payment_term.xml
grindtildeath: Actually the module removes the `sequence` field which could be of some use. I don't think it's with intent ?
The reason looks simply to be, that this view did not exist prior to v11, but as it now does, IMO we should respect the inheritance.

account_early_payment_discount/views/account_payment_term.xml
grindtildeath: No, it's not, but it makes no harm and it's better to have a name when you have to select a record for any field on the GUI.

asaunier:
@grindtildeath OK with your arguments. I have made a fixup commit to take them into account.

_Edit: squashed the fixup commits_

grindtildeath (approved):
Thank you @asaunier :+1:

asaunier:
Fixup commits have been squashed. This PR can be merged if no other objection is raised. Thanks :)

rafaelbn:
Hello! Nobody answer to https://github.com/OCA/account-payment/pull/184#issuecomment-383139170 of @feketemihai about https://github.com/OCA/account-payment/pull/174 :thinking: any case thank! I will try it! :smile:

PR #189, opened 2018-05-18
[IMP] account_check_printing_report_base: Compute due date through move lines

pedrobaeza:
When you have confirmed the invoice and change the due date in the move lines instead of the invoice, the due date that appear on the checks is still the invoice one.

With this commit, you get that one, but falling back to the invoice one if not present on move lines (which is rare).

cc @Tecnativa


LoisRForgeFlow (approved):
LGTM :+1:

PR #191, opened 2018-06-18
[10.0] [IMP] Add store=True to payment_term_id

giusygiallara93:
I added store = True to the payment_term_id field to be able to insert it in the filters and in the group_by.


pedrobaeza:
I don't agree with this, as the important thing is the due date, not the payment term. And this will provoke a big computation for storing the field.

giusygiallara93:
I made these changes because so you have the possibility to filter and group for the payment term.

PR #193, opened 2018-06-25
Fix ERROR on the usage of computed domain

lfr-opennet:
The error "Error: TypeError: 'float' object is unsubscriptable" appears when trying to change the journal_in on the wizard view.

### DIFF
--- a/account_cash_invoice/wizard/cash_invoice_in.xml
+++ b/account_cash_invoice/wizard/cash_invoice_in.xml
@@ -8,11 +8,15 @@
             <form string="Pay supplier invoice or customer refund">
                 <group>
                     <field name="company_id" invisible="1"/>
-                    <field name="journal_ids" invisible="1"/>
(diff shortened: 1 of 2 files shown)


account_cash_invoice/wizard/cash_invoice_in.xml
```diff
@@ -8,11 +8,15 @@
             <form string="Pay supplier invoice or customer refund">
                 <group>
                     <field name="company_id" invisible="1"/>
-                    <field name="journal_ids" invisible="1"/>
+                    <field name="journal_ids" invisible="1">
+                        <tree>
+                            <field name="id"/>
```
pedrobaeza, line 13: Is this really needed?

lfr-opennet: On the cash_invoice_id view is probably not needed but at least if there is more than 1 journal to select, it will not crash an error.

PR #196, opened 2018-08-17
[FIX] Fix for contibutors

mgosai:


max3903 (approved):

account_payment_credit_card/__manifest__.py
```diff
@@ -1,14 +1,14 @@
...
 {
     "name": "Credit Card Payments",
     "summary": "Add support for credit card payments",
     "version": "11.0.1.0.0",
     "license": "AGPL-3",
-    "author": "Ursa Information Systems, Odoo Community Association (OCA)",
+    "author": "Open Source Integrators, Odoo Community Association (OCA)",
     "category": "Accounting & Finance",
-    "maintainer": "Ursa Information Systems",
+    "maintainer": "Open Source Integrators",
```
pedrobaeza: This key is incorrect. You should put here a GitHub login or remove it.

max3903: @pedrobaeza The key for Github logins is `maintainers` ;)
https://github.com/OCA/mis-builder/blob/10.0/mis_builder/__manifest__.py#L40

https://github.com/OCA/OCB/blob/11.0/odoo/addons/base/module/module.py#L254

pedrobaeza: OK, then what is the sense of putting a redundant information in a non used key?

max3903: @pedrobaeza `maintainer` comes from Odoo and contains a company name. `maintainers` was added by @sbidoul and contains Github logins.

pedrobaeza: OK, I won't insist more, but I still think it's a useless key not shown anywhere.

max3903:
@pedrobaeza If that's all it takes to get your approval, happy to remove it :)

pedrobaeza (approved):
Hehe, thanks for the consideration.

PR #197, opened 2018-08-27
[11.0][MIG] account_due_list_aging_comments

MiquelRForgeFlow:
Standard migration.


pedrobaeza:
Without plurals, please :wink: Isn't better to use `account_credit_control` for this?

MiquelRForgeFlow:
I don't know much about the `account_credit_control` module, but I see that has many features, that maybe the people that use the `account_due_list_aging_comment` module don't want.

pedrobaeza:
Well, you are having a lot of modules for credit control (this one, the aging days, and so on), that are treated in a more flow oriented way in that module, so it deserves you to check it (tell me if you need a quick tour) and maybe replace them.

MiquelRForgeFlow:
Well, the quick tour would be really nice 😊

JordiBForgeFlow (approved):

JordiBForgeFlow:
Quick merge. This is a trivial migration.

PR #204, opened 2018-10-23
[10.0][IMP] account_payment_show_invoice: Add invoice on move line level

rousseldenis:
* Add invoice on move level
* Fix interface

### DIFF
--- a/account_payment_show_invoice/__manifest__.py
+++ b/account_payment_show_invoice/__manifest__.py
@@ -5,7 +5,7 @@
     "name": "Account Payment Show Invoice",
     "summary": "Extends the tree view of payments to show the paid invoices "
                "related to the payments using the vendor reference by default",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.1.0",
     "category": "Account-payment",
     "website": "https://odoo-community.org/",
     "author": "Eficent, Odoo Community Association (OCA)",
@@ -15,5 +15,6 @@
     "depends": ["account"],
     "data": [
         "views/account_payment_view.xml",
+        "views/account_move_line.xml",
     ],
 }
--- a/account_payment_show_invoice/models/account_move_line.py
+++ b/account_payment_show_invoice/models/account_move_line.py
@@ -1,0 +1,16 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class AccountMoveLine(models.Model):
+
+    _inherit = 'account.move.line'
+
+    payment_invoice_ids = fields.Many2many(
+        string="Payment Invoices",
+        related='payment_id.invoice_ids',
+        readonly=True,
+    )
--- a/account_payment_show_invoice/views/account_move_line.xml
+++ b/account_payment_show_invoice/views/account_move_line.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2018 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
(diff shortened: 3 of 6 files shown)


rousseldenis:
@lreficent What do you think of this ?
cc @acsone-odoo-team

Cedric-Pigeon (approved):
code review

account_payment_show_invoice/__manifest__.py
```diff
@@ -5,7 +5,7 @@
     "name": "Account Payment Show Invoice",
     "summary": "Extends the tree view of payments to show the paid invoices "
                "related to the payments using the vendor reference by default",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.1.0",
     "category": "Account-payment",
     "website": "https://odoo-community.org/",
     "author": "Eficent, Odoo Community Association (OCA)",
```
ThomasBinsfeld, line 11: @rousseldenis The new field does not appear. You forgot to add the view file in the manifest.

ThomasBinsfeld (changes requested):

account_payment_show_invoice/__manifest__.py
rousseldenis: Done

ThomasBinsfeld (approved):
Code and functional review :+1:

Cedric-Pigeon:
@OCA/accounting-maintainers I think this one could be merged... Thanks

account_payment_show_invoice/views/account_move_line.xml
```diff
@@ -0,0 +1,25 @@
...
+                </xpath>
+            </field>
+    </record>
+    <record id="view_move_line_tree" model="ir.ui.view">
+            <field name="name">account.move.line.tree (in account_payment_show_invoice)</field>
+            <field name="model">account.move.line</field>
+            <field name="inherit_id" ref="account.view_move_line_tree"/>
+            <field name="arch" type="xml">
+                <field name="move_id" position="after">
+                    <field name="payment_invoice_ids"/>
```
eLBati: @rousseldenis adding a not stored computed field to `account.move.line` list could be dangerous for performance

rousseldenis: @eLBati They are direct fields that are both stored (payment_id.invoice_ids). So, I think it should be acceptable (but depends on ORM - which has been improved on related fields)

account_payment_show_invoice/models/account_move_line.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class AccountMoveLine(models.Model):
+
+    _inherit = 'account.move.line'
+
+    payment_invoice_ids = fields.Many2many(
+        string="Payment Invoices",
+        related='payment_id.invoice_ids',
+        readonly=True,
```
eLBati, line 15: @rousseldenis do you see problems with adding this
```suggestion
        readonly=True,
        store=True
```
?

rousseldenis:
@lreficent @mreficent Concerning this module, there is a bug in interface if tree view is in editable bottom mode. invoice_ids field in account.payment tree should be set with widget "many2many_tags". This PR should be modified too. I will do the fix in this in another commit if you agree.

ThomasBinsfeld (approved):

account_payment_show_invoice/views/account_move_line.xml
rousseldenis: @eLBati In production for months and no performance problem. Mutli-company, thousands of moves. Could you update your review?

eLBati: :+1:

eLBati:
@rousseldenis I think the 2 commits should be squashed, as concerning the same change. Do you prefer to keep them separate?

Moreover, are these changes also applicable to v11?

Thanks

rousseldenis:
@eLBati I'd prefer keeping them as it will facilitate feedback on that problem.

eLBati:
@rousseldenis ok. What about v11?

rousseldenis:
@eLBati Didn't test in v11. But I think this should be the case too.

PR #210, opened 2018-11-16
FIX payment mode on account move line

### LINKED ISSUE #208: [11.0] account_due_list_payment_mode & account_payment_partner issue
Hi,

module account_due_list_payment_mode create payment_mode_id field on account.move.line as related='invoice_id.payment_mode_id', so it's not editable.

Module account_payment_partner create other field with same name, but not as related, this field is editable and is set on invoice validation.

For user is usefull to get this field editable on account_due_list.

Module account_due_list_payment_mode depends account_payment_partner, I think that could be better to don't redefine this field and allow to edit on account due list.

angelmoya:
Fixes #208


pedrobaeza (approved):

ljimenezsidoo (approved):
:+1:

PR #215, opened 2019-01-28
[12.0] [MIG] Account Payment Return Import

ThomasBinsfeld:


account_payment_return_import/tests/test_import_payment_return.py
```diff
@@ -12,6 +12,7 @@ def setUp(self):
         self.company = self.env.ref('base.main_company')
         self.acc_number = 'NL77ABNA0574908765'
         self.acc_bank = self.env['res.partner.bank'].create({
+            'partner_id': self.env.ref('base.res_partner_1').id,
```
carlosdauden: Can you change with this:
`'partner_id': self.company.partner_id.id`
for it hasn't demo data dependences

Should not the company be?:
`'company_id': self.company.id`

carlosdauden (approved):

account_payment_return_import/tests/test_import_payment_return.py
ThomasBinsfeld: Done

carlosdauden: Thanks

pedrobaeza:
You can rebase now

ThomasBinsfeld:
@pedrobaeza Done

account_payment_return_import/models/payment_return.py
```diff
@@ -17,5 +17,5 @@ class PaymentReturnLine(models.Model):
     _sql_constraints = [
         ('unique_import_id',
          'unique (unique_import_id)',
-         'A bank account transactions can be imported only once !')
+         'A bank account transaction can be imported only once !')
```
pedrobaeza: Now that you touch this, proper message should be:
```A payment return transaction can be imported only once !```

File: account_payment_return_import/readme/CONTRIBUTORS.rst [hunk 3]
```diff
@@ -0,0 +1,4 @@
+* Carlos Dauden <pedro.baeza@tecnativa.com>
+* Pedro M. Baeza <pedro.baeza@tecnativa.com>
+* David Vidal <david.vidal@tecnativa.com>
+* Luis M. Ontalba <luis.martinez@tecnativa.com>
```
pedrobaeza, line 4: You can add yourself here

account_payment_return_import/wizard/payment_return_import_view.xml
```diff
@@ -6,7 +6,13 @@
         <field name="model">payment.return.import</field>
         <field name="priority">1</field>
         <field name="arch" type="xml">
-            <form string="Import Payment Returns">
+            <form>
+                <div name="info_alert" class="alert alert-info" role="alert">
+                    How to import your payment return :
```
pedrobaeza: Is this going to be translatable this way? I think you have to enclose it in a tag (`<span>` for example). Previous method was OK. Why change it?

pedrobaeza (changes requested):

account_payment_return_import/wizard/payment_return_import_view.xml
ThomasBinsfeld: @pedrobaeza Odoo 12 does not accept `<label>` without `for` attribute.

`<label string= "1. Download your Payment Return file from your bank website."/>` doesn't work (error while validating the view).
`<label string= "1. Download your Payment Return file from your bank website." for="date_file"/>` would work just fine but is confusing. 

I'll check the translation of my way to do.

pedrobaeza: Right, but you can put the same `for` for all. The problem is that changing this, you will waste all the done translations.

ThomasBinsfeld: @pedrobaeza Well, a migration is the best opportunity to get rid of an ugly code/interface. We'll only lost the translation of 4 sentences. Moreover, it will be far easier to translate them by putting all 4 sentences in a div instead of separate labels.

pedrobaeza: OK, go ahead if the translation works.

ThomasBinsfeld:
@pedrobaeza Code updated.

account_payment_return_import/tests/test_import_base.py
```diff
@@ -0,0 +1,102 @@
...
+    """Run test to import payment return import."""
+
+    def setUp(self):
+        super(TestImportBase, self).setUp()
+        self.company = self.env.ref('base.main_company')
+        self.acc_number = 'NL77ABNA0574908765'
+        self.acc_bank = self.env['res.partner.bank'].create({
+            'acc_number': self.acc_number,
+            'bank_name': 'TEST BANK',
+            'company_id': self.company.partner_id.id,
```
Nikul-Chaudhary: `self.company.partner_id.id,` ??

Nikul-Chaudhary (commented):
Improve code

account_payment_return_import/tests/test_import_base.py
ThomasBinsfeld: Fixed

pedrobaeza (approved):

PR #224, opened 2019-02-19
[IMP] account_check_printing: Allow to use other models

etobella:
This way, we could use account_check_printing_report_base with other kinds of elements, like `account.payment.order` from bank-payment

### DIFF
--- a/account_check_printing_report_base/report/check_print.py
+++ b/account_check_printing_report_base/report/check_print.py
@@ -4,7 +4,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 import time
-from odoo import api, exceptions, models, _
+from odoo import api, models
 from odoo.tools import float_is_zero
 
 
@@ -20,6 +20,8 @@ class ReportCheckPrint(models.AbstractModel):
 
     @api.multi
     def get_paid_lines(self, payments):
+        if self.env.context.get('active_model') != 'account.payment':
+            return {}
         lines = {}
         for payment in payments:
             lines[payment.id] = []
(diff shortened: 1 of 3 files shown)


JordiBForgeFlow (approved):

AaronHForgeFlow (approved):
:+1: LGTM

account_check_printing_report_base/report/check_print.py
```diff
@@ -92,16 +94,15 @@ def get_paid_lines(self, payments):
...
             'doc_ids': docids,
-            'doc_model': 'account.payment',
-            'docs': payments,
+            'doc_model': models,
+            'docs': objects,
             'time': time,
             'fill_stars': self.fill_stars,
             'paid_lines': paid_lines
         }
-        if self.env.user.company_id.check_layout_id:
```
pedrobaeza, line 105: I think this is still needed

etobella: If the layout is not defined, the following error wil be raised: https://github.com/odoo/odoo/blob/11.0/addons/account_check_printing/models/account_payment.py#L120

Here, you are already on a report, so I think it is not necessary to check the layout.

pedrobaeza: Is that method called always?

etobella: Not necessarily, but If you have selected a layout in order to print a check, why should it fail if your company (not the company of the payment) have no layout for the checks defined?

pedrobaeza: But how do you select that layout? For now, there's no other option.

etobella: Yes, but the layout is selected earlier on https://github.com/OCA/account-payment/blob/11.0/account_check_printing_report_base/models/account_payment.py#L30. If no layout is selected, it will raise the error I showed.

pedrobaeza: OK, but next time please don't make these extra changes without advertising, as this is not the aim of this PR.

Swapnesh-SerpentCS:
Hello All,
It seems Issue after this merge, active_model is not available in the context (Issue https://github.com/OCA/account-payment/issues/230) (or maybe I am missing something here)

Added Fix in https://github.com/OCA/account-payment/pull/267

Cc: @max3903

Issue #226, opened 2019-02-25
[12.0] [RFC] Account Payment Return Import SEPA PAIN: reason codes


ThomasBinsfeld:
The [ISO20022 documentation](https://www.iso20022.org/external_code_list.page) lists 225 return codes while the [OCA addon](https://github.com/OCA/account-payment/blob/12.0/account_payment_return_import_sepa_pain/data/unpaid_reason_codes.xml) lists 20 codes. Shouldn't all these missing codes be added in the OCA addon?

pedrobaeza:
Well, I don't remember well the source of this, but 225 can be too much, isn't it? Should we put all of them or only the most use ones?

sbidoul:
Question is: how to know which are the most used ones? I propose to add them all. 

If at some point people combine manual entry of return with PAIN import, one possibility could be adding active flag on return codes so they don't appear in the drop down list, while the import code could still use them.

ThomasBinsfeld:
Would a XML file the best way to embed the 225 return reasons in the module (in terms of maintainability) ? I don't know if it can be done with a CSV file (no `noupdate` option).

pedrobaeza:
As this is standard, it can be tolerable to be put as CSV noupdate=0, as you shouldn't touch here. If you want, then create others. What do you think?

ThomasBinsfeld:
Done in 12.0, I close this issue.

PR #228, opened 2019-02-25
[12.0] [ADD] Account Payment Return Import SEPA PAIN: missing return reasons

ThomasBinsfeld:
From #226


sbidoul (approved):

pedrobaeza (approved):

pedrobaeza:
I'm thinking that we should have `_rec_name = complete_name` or `name_get` with "[Code] Name" for not loosing this codification in UI. What do you think?

sbidoul:
> [Code] Name

Yes good idea. @ThomasBinsfeld can you do it while you are at it?

ThomasBinsfeld:
@pedrobaeza @sbejaoui Voilà

account_payment_return/models/payment_return_reason.py
```diff
@@ -20,3 +20,9 @@ def name_search(self, name, args=None, operator='ilike', limit=100):
         if not recs:
             recs = self.search([('name', operator, name)] + args, limit=limit)
         return recs.name_get()
+
+    @api.multi
+    @api.depends('code', 'name')
```
pedrobaeza: This is not needed, as it's not a compute field

account_payment_return/models/payment_return_reason.py
```diff
@@ -20,3 +20,9 @@ def name_search(self, name, args=None, operator='ilike', limit=100):
         if not recs:
             recs = self.search([('name', operator, name)] + args, limit=limit)
         return recs.name_get()
+
+    @api.multi
```
pedrobaeza, line 24: This is neither, as it's the default.

account_payment_return/models/payment_return_reason.py
ThomasBinsfeld: Right, I removed it.

account_payment_return/models/payment_return_reason.py
ThomasBinsfeld: Yes, but I prefer keep it explicit.

Issue #227, opened 2019-02-25
[12.0] [RFC] Account Payment Return Import SEPA PAIN: NARR reason


ThomasBinsfeld:
In a SEPA payment status report, the reason code can be [NARR](https://www.iso20022.org/external_code_list.page), meaning the description doesn't match any reason code but is described in the `AddtlInf` tag. 

Shouldn't the OCA addon manage these narrative reasons?

pedrobaeza:
Yes, but it supposes to add a new field for free text. Maybe on the main module?

sbidoul:
+1 for a free text field on return lines in the main module.

ThomasBinsfeld:
Done in 12.0, I close this issue.

PR #229, opened 2019-02-26
[12.0] [ADD] Payment Return Import SEPA: reason additional info

ThomasBinsfeld:
New free text field on return lines: additional information on return reason.
This field is populated in account_payment_return_import_sepa_pain with the content of the AddtlInf tag.
This is useful when the return reason code is NARR, meaning the reason description does not match any known code but is in this tag.

---------------------
From #227

### DIFF
--- a/account_payment_return/models/payment_return.py
+++ b/account_payment_return/models/payment_return.py
@@ -235,6 +235,10 @@ class PaymentReturnLine(models.Model):
         oldname="reason",
         string='Return reason',
     )
+    reason_additional_information = fields.Char(
+        string="Return reason (info)",
+        help="Additional information on return reason.",
+    )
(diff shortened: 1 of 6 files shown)


pedrobaeza (commented):
My only concern here is the length of the label of the new field, as it will take a lot of space in the tree. Can we look for a shorter label (maybe put a shorter one in the `string` of the XML?).

It's also curious that you add the field in the SEPA, but doesn't test for the value in the tests.

ThomasBinsfeld:
@pedrobaeza I fixed the test and changed the label (it fits on one line now).

pedrobaeza (approved):

sbidoul (approved):

PR #234, opened 2019-03-05
[12.0] account_due_list: prevent to create or delete move lines, as not a use case for this module

eLBati:
Improve summary

### DIFF
--- a/account_due_list/__manifest__.py
+++ b/account_due_list/__manifest__.py
@@ -6,7 +6,7 @@
     'category': 'Generic Modules/Payment',
     'development_status': 'Beta',
     'author': 'Odoo Community Association (OCA)',
-    'summary': 'Check printing commons',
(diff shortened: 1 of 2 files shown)


tbaden (approved):
code review: lgtm and sounds reasonable to prevent it

account_due_list/__manifest__.py
```diff
@@ -6,7 +6,7 @@
     'category': 'Generic Modules/Payment',
     'development_status': 'Beta',
     'author': 'Odoo Community Association (OCA)',
-    'summary': 'Check printing commons',
+    'summary': 'List of open credits and debts, with due date',
```
pedrobaeza: s/debts/debits

eLBati: @pedrobaeza done

pedrobaeza (approved):

PR #237, opened 2019-03-13
[12.0] [IMP] Payment Return: optional automatic reconciliation

ThomasBinsfeld:
A new option on the journal allowing to try an automatic reconciliation of the payment return. It only makes sense when using transfer accounts. I the reconciliation failed for some reason, a warning message is displayed on the form view.

### DIFF
--- a/account_payment_return/models/account_journal.py
+++ b/account_payment_return/models/account_journal.py
@@ -14,3 +14,10 @@ class AccountJournal(models.Model):
         comodel_name="res.partner", string="Default Charges Partner",
         domain=[('supplier', '=', True)], help='Default partner for '
                                                'charge expenses')
+    return_auto_reconcile = fields.Boolean(
+        string="Reconcile payment returns",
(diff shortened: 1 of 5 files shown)


kittiu:
@ThomasBinsfeld thanks for the contribution. Should we use [IMP] for branch name and commit message? Do we use [ADD] only if we add new files?

ThomasBinsfeld:
> @ThomasBinsfeld thanks for the contribution. Should we use [IMP] for branch name and commit message? Do we use [ADD] only if we add new files?

@kittiu Yes indeed, i changed the commit message and PR title.

pedrobaeza:
`[ADD]` is only for new modules indeed

chrisdec (approved):
Functional tests

sbejaoui (approved):
Code review

ThomasBinsfeld:
@pedrobaeza  This PR is ready to be merged

PR #242, opened 2019-03-29
[12.0][MIG] account_payment_show_invoice

kittiu:


account_payment_show_invoice/views/account_payment_view.xml
```diff
@@ -0,0 +1,30 @@
...
+        </field>
+    </record>
+
+    <record id="view_account_payment_tree" model="ir.ui.view">
+        <field name="name">account.supplier.payment.tree - show invoice</field>
+        <field name="model">account.payment</field>
+        <field name="inherit_id" ref="account.view_account_payment_tree"/>
+        <field name="arch" type="xml">
+            <field name="partner_id" position="after">
+                <field name="invoice_vendor_references"/>
```
kittiu, line 25: @lreficent , I change from **invoice_ids** to **invoice_vendor_references**. (not sure invoice_ids is your intention?)

account_payment_show_invoice/models/account_payment.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2017 Eficent Business and IT Consulting Services, S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class AccountPayment(models.Model):
+    _inherit = 'account.payment'
+
+    @api.multi
```
kittiu, line 10: @lreficent I change from api.one -> api.multi. In the past, I heard that we should always use api.multi since api.one have performance issue.
But I also see many code using api.one, not sure the performance was fixed. Please suggest.

account_payment_show_invoice/models/account_payment.py
```diff
@@ -0,0 +1,19 @@
...
+    def _compute_invoice_vendor_references(self):
```
ntodorova, line 11: i think that it is a good practice to include ensure_one() if you change the api from one to multi.

account_payment_show_invoice/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from . import test_payment_ref_invoice
```
ntodorova, line 3: dont't we need a license text here?

account_payment_show_invoice/views/account_payment_view.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2017 Eficent
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+    <record id="view_account_supplier_payment_tree" model="ir.ui.view">
+        <field name="name">account.supplier.payment.tree - show invoice</field>
```
ntodorova, line 8: Is changing the name required?

account_payment_show_invoice/views/account_payment_view.xml
```diff
@@ -0,0 +1,30 @@
...
+               ref="account.view_account_supplier_payment_tree"/>
+        <field name="arch" type="xml">
+            <field name="partner_id" position="after">
+                <field name="invoice_vendor_references"/>
+            </field>
+        </field>
+    </record>
+
+    <record id="view_account_payment_tree" model="ir.ui.view">
+        <field name="name">account.supplier.payment.tree - show invoice</field>
```
ntodorova, line 20: same as above.

account_payment_show_invoice/static/description/index.html
```diff
@@ -0,0 +1,431 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
ntodorova, line 1: Is this file changed, for me it looks like the one in v11?

account_payment_show_invoice/models/account_payment.py
```diff
@@ -0,0 +1,19 @@
...
+        for payment in self:
+            ref = payment.invoice_ids.mapped(lambda x: x.reference or x.number)
+            ref.sort()
+            payment.invoice_vendor_references = ', '.join(ref)
```
ntodorova, line 15: I like this function implementation much more than in v11, good job!

ntodorova (changes requested):

ntodorova:
It looks good, even better in v11. I just have couple of questions before I approve it.

account_payment_show_invoice/models/account_payment.py
kittiu: Here, I think the reason to use multi is oppose of using ensure_one() here isn't it? We use multi because we want to compute once for all records.
Please correct me if I am wrong.

account_payment_show_invoice/views/account_payment_view.xml
kittiu: Can you suggest please? I am not sure what you means (I just migrate from original)

account_payment_show_invoice/static/description/index.html
kittiu: This one, I did follow the instruction by using **oca-gen-addon-readme**

account_payment_show_invoice/models/account_payment.py
kittiu: Thank you :)

account_payment_show_invoice/tests/__init__.py
kittiu: I will add it.

account_payment_show_invoice/views/account_payment_view.xml
ntodorova: In the original is "account.supplier.payment.tree" and here it is "account.supplier.payment.tree - show invoice", I am just not sure if it should be changed or not : )

account_payment_show_invoice/models/account_payment.py
ntodorova: When I double think, should that not be @api.depends('invoice_ids')?
Because the field should be changed in case the 'invoice_ids' are changed.

account_payment_show_invoice/views/account_payment_view.xml
kittiu: Yes. I just check from v11 also. And it also has the - show invoice

ntodorova: True, I am sorry. I have overseen it. That here all is fine.

ntodorova (commented):
ok, so only the api question is left.

kittiu:
> ok, so only the api question is left.

Note really, because this is not **store=True** field. So, it is not necessary (it will be recomputed every time it is called)

account_payment_show_invoice/README.rst
```diff
@@ -0,0 +1,84 @@
...
+.. contents::
+   :local:
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to 'Invoicing > Vendors > Payments' or to 'Invoicing > Customers >
+   Payments'
```
ntodorova, line 42: I did a function test and did not found the menus mentioned here. Can you please check this too?

account_payment_show_invoice/models/account_payment.py
ntodorova: ok, you are right, after we discussed it I agree that @api.multi without ensure_one is fine. : ))

account_payment_show_invoice/README.rst
kittiu: image

ntodorova (approved):
LGTM

cfreyse (approved):

tbaden (changes requested):
https://github.com/OCA/account-payment/blob/a451bb6aa9689efa592789f649045f8d6d724ca5/account_payment_show_invoice/__manifest__.py#L9

please change the website in the manifest to: https://github.com/OCA/account-payment

kittiu:
> [account-payment/account_payment_show_invoice/__manifest__.py](https://github.com/OCA/account-payment/blob/a451bb6aa9689efa592789f649045f8d6d724ca5/account_payment_show_invoice/__manifest__.py#L9)
> 
> Line 9 in [a451bb6](/OCA/account-payment/commit/a451bb6aa9689efa592789f649045f8d6d724ca5)
> 
>  "website": "https://odoo-community.org/", 
> please change the website in the manifest to: https://github.com/OCA/account-payment

Done.

tbaden (approved):
Code Review

PR #243, opened 2019-04-01
[10.0][IMP] account_cash_discount_base: Allows to enter manual discount

rousseldenis:
Adds a parameter on Company level to authorize or not the manual cash discount.
Adds a parameter on invoice level to enable manual discount field.
Adds the manual cash discount on invoice level.

### DIFF
--- a/account_cash_discount_base/README.rst
+++ b/account_cash_discount_base/README.rst
@@ -21,6 +21,9 @@ is still computed at invoice's validation.
 You can decide to calculate the discount amount based on the total price taxes
 included or not.
 
+You can also define a manual amount on invoice level if your company configuration
+allows it.
+
 
 Configuration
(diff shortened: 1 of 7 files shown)


tbaden (commented):
https://github.com/OCA/account-payment/blob/eead79027cb95b5e12da77f96c7d4d02ae0c3f08/account_cash_discount_base/tests/test_account_cash_discount_base.py#L121-L136

is it really necessary to test percentage discount in manual_cash_discount test?
and if it is, than there should be two tests for untaxed and total like these:

https://github.com/OCA/account-payment/blob/eead79027cb95b5e12da77f96c7d4d02ae0c3f08/account_cash_discount_base/tests/test_account_cash_discount_base.py#L52-L74

PR #256, opened 2019-05-22
[12.0][MIG] account_due_list_payment_mode

MiquelRForgeFlow:
Standard migration.


account_due_list_payment_mode/README.rst
```diff
@@ -27,7 +27,7 @@ to check "Show full accounting features" in user permission.
 Usage
 =====
 
-#. Go to *Invoicing > Adviser > Payments and due list* and you'll see the new
+#. Go to *Invoicing > Adviser > Payments and due list* and you will see the new
```
pedrobaeza, line 29: Why changing this? It was correct previously.

account_due_list_payment_mode/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,4 @@
+* Obertix Free Solutions <http://obertix.net>
+* Sergio Teruel <sergio.teruel@tecnativa.com>
+* Albert De La Fuente <info@haevas.com>
+* Vicent Cubells <vicent.cubells@tecnativa.com>
```
pedrobaeza, line 4: Add yourself here

account_due_list_payment_mode/readme/INSTALL.rst
```diff
@@ -0,0 +1,4 @@
+This module depends on:
+
+* account_due_list
```
pedrobaeza: No need to indicate this one, as it's on the same repo.

account_due_list_payment_mode/readme/INSTALL.rst
```diff
@@ -0,0 +1,4 @@
...
+* account_payment_partner
```
pedrobaeza: Indicate that this one is on the repo OCA/bank-payment.

pedrobaeza (changes requested):

carlosdauden (approved):

account_due_list_payment_mode/README.rst
MiquelRForgeFlow: I know. But I think it's better use formal writing. In English, the use of the apostrophe is more colloquial or casual.

account_due_list_payment_mode/readme/CONTRIBUTORS.rst
MiquelRForgeFlow: why? is it necessary? my contribution it's just a simple migration, only changing version number. 🤔 But ok.

pedrobaeza: Well, if you don't want... but I think each contribution deserves to be listed.

account_due_list_payment_mode/README.rst
pedrobaeza: Yes, I can agree. As always is for not changing too much diff, but this can be kept indeed.

account_due_list_payment_mode/readme/CONTRIBUTORS.rst
MiquelRForgeFlow: ok, I included myself.

pedrobaeza (approved):

PR #259, opened 2019-05-25
[FIX] account_check_printing_report_base: invoice with multiple payments

### LINKED ISSUE #258: [10.0] `account_check_printing_report_base` check reports report incorrect amount in "Payment"  column
For invoices with multiple payments, amounts listed in "Payment" column are the same as amounts in the "Original Amount".

The issue is that the report is aggregating amounts from all of the invoice's payments rather than selecting the ones in the report.

brendanmasson:
Fix check printing to work with invoices that have multiple payments.
The amount in the "Payment" column should only include amounts related
to the selected payment.

Fixes #258.

### DIFF
--- a/account_check_printing_report_base/report/check_print.py
+++ b/account_check_printing_report_base/report/check_print.py
@@ -47,7 +47,10 @@ class ReportCheckPrint(models.AbstractModel):
                 if invoice.type == 'out_refund':
                     line['amount_total'] *= -1
                 total_amount_to_show = 0.0
-                for pay in invoice.payment_move_line_ids:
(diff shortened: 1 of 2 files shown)


Swapnesh-SerpentCS:
I think this has been handled in https://github.com/OCA/account-payment/pull/261
Cc @mreficent  @jbeficent

MiquelRForgeFlow:
conflicts

Swapnesh-SerpentCS:
I think we can close this as https://github.com/OCA/account-payment/pull/261 works for all AMLs

PR #268, opened 2019-06-13
[12.0] [IMP] Payment Return Import ISO 20022: manage multiple notification, entry and details nodes

ThomasBinsfeld:
There may be more than one details node in a single entry node.
They must create one return line each.
The return amount must be taken in the details node instead of the entry node.


sbidoul (approved):
(code review)

chrisdec (approved):
Functional test

PR #270, opened 2019-06-14
[12.0] [IMP] Payment Return: usability

ThomasBinsfeld:
* New total field on payment returns
* Total footers
* Display the bank account from the imported file on the payment orders (useful when using transfer accounts)

### DIFF
--- a/account_payment_return/views/payment_return_view.xml
+++ b/account_payment_return/views/payment_return_view.xml
@@ -79,6 +79,27 @@
         </field>
     </record>
 
+    <record model="ir.ui.view" id="payment_return_search_view">
+        <field name="name">payment.return.search</field>
+        <field name="model">payment.return</field>
+        <field name="type">search</field>
+        <field name="arch" type="xml">
+            <search>
+                <field name="name"/>
+                <group expand="0" string="Group By">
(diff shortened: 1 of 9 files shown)


chrisdec (approved):
Functional Test

account_payment_return/views/payment_return_view.xml
```diff
@@ -43,8 +44,8 @@
                                                     ('account_id.internal_type', '=', 'receivable'),
                                                     ('reconciled', '=', True)]"
                                             />
-                                    <field name="amount"/>
-                                    <field name="expense_amount" />
+                                    <field name="amount" sum="Total amount"/>
+                                    <field name="expense_amount" sum="Total expense amount"/>
```
sbidoul: The string for `expense_amount` is `Charges Amount`, so it's better to use the same vocabulary for the total.

sbidoul (changes requested):
A small vocabulary issue, otherwise looks good (code review).

account_payment_return/views/payment_return_view.xml
ThomasBinsfeld: Done, thanks!

sbidoul (approved):

PR #271, opened 2019-06-17
[10.0][IMP] account_cash_discount_payment: Don't be too restrictive w…

rousseldenis:
…ith constrains


rousseldenis:
@sbidoul @chrisdec @benwillig

sbidoul (approved):
@rousseldenis looks good. A word in the commit message explaining *why* this change is necessary would be nice.

PR #273, opened 2019-07-02
[12.0][FIX] - payment return reason csv file missing in manifest

sbejaoui:
This PR fixes:
- Missing quotes in L216 and 217 to escape comma used in name
- Payment return reason csv file missing in manifest

### DIFF
--- a/account_payment_return_import_iso20022/__manifest__.py
+++ b/account_payment_return_import_iso20022/__manifest__.py
@@ -20,5 +20,6 @@
         'account_payment_order',
     ],
     'data': [
+        "data/payment.return.reason.csv"
     ],
 }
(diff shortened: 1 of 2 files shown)


sbejaoui:
@ThomasBinsfeld, can you check this please

ThomasBinsfeld (approved):

pedrobaeza:
Not sure if we already discussed this, but isn't a lot of reasons to be loaded?

ThomasBinsfeld:
@pedrobaeza It's all the reasons defined by ISO20022, assuring that all return codes we receive will be well interpreted by the module.

pedrobaeza:
Yeah, I know, but maybe it's overwhelmed. Let's see what people says.

pedrobaeza (approved):

PR #241, opened 2019-03-28
[12.0][MIG] account_check_printing_report_base

Abranes:
Standard migration to 12.0

Depends on:
* [x] OCA/reporting-engine/pull/278

PR #278, opened 2019-08-01
[12.0][FIX] account_check_printing_report_base: Report format

Abranes:
As @aheficent says in https://github.com/OCA/account-payment/pull/241#issuecomment-502039004 the report is displaying bad.

Before changes:
old_report

After changes:
now_report

### DIFF
--- a/account_check_printing_report_base/views/report_check_base.xml
+++ b/account_check_printing_report_base/views/report_check_base.xml
@@ -5,84 +5,80 @@
         <t t-call="web.html_container">
             <div class="header"/>
             <div class="article">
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

AaronHForgeFlow (approved):
It prints correctly to me now. For some reason, before this change, it was printed correct for other people. Could it be something related to wkthmtopdf version?

Abranes:
I don't know. I tried in local and in Runbot and in both places it was printed wrong. I use `wkhtmltopdf 0.12.5`

fuentes73 (approved):

PR #285, opened 2019-10-01
 [13.0] [MIG] account_payment_credit_card

Nikul-Chaudhary:
Credit Card Payments
====================
This module provides a way to record Credit Card purchases and to pay them the same way regular purchases are recorded (via Vendor Bills) and paid (via Vendor Payments).

By default, if vendor bills are used to record credit card purchases, the liability (AP) from the supplier/vendor must be moved to the credit card company via a manual journal entry.

Because a journal entry contains less information than a vendor bill, there may be a loss of information that affects book keeping accuracy and the ability to properly reconcile the AP amount when the credit card statement arrives and needs to be settled.
This method also requires an additional step not needed when a vendor bill is paid via bank or cash.

This module automates the creation of the manual journal entry required to move the AP to the credit card company, retains all information about the purchase in an invoice document, and leverages the standard purchase and payment workflow already in place for bank and cash payments.

It supports as many Credit Cards as are needed, configured in the same way an additional payment method would be (i.e. via the creation of a new journal).

To settle a Credit Card statement, the regular workflow to record a vendor payment is used – allowing the removal of charges not included in the statement.
Non-purchase transactions like fees for annual membership, balance transfers, cash advances and foreign transactions; as well as charges for late payments and returned checks; can be entered either as Vendor Invoices or manual Journal Entries as users elect.

Both of these methods will allow these items to be settled when making a payment to the credit card company.

This module also supports payment cancellation and re-entry (in the case a mistaken amount is entered) as well as refunds (where the purchase is returned and a credit from the credit card company will be issued).


Nikul-Chaudhary:
@bodedra @max3903 can you please review it?

Nikul-Chaudhary:
@bodedra @max3903 If you don't mind, can you please review it?

account_payment_credit_card/models/account_move.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2017 Open Source Integrators <http://www.opensourceintegrators.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def post(self, invoice=False):
```
kongrattapong:
```suggestion
    def post(self):
```
In Odoo13 ```def post``` not require "invoice"

account_payment_credit_card/models/account_move.py
```diff
@@ -0,0 +1,40 @@
...
+                            "move_id": move_line.move_id.id,
+                            "date": move_line.date,
+                        }
+                    )
+                # Check result list
+                if result:
+                    # Create new move lines
+                    for vals in result:
+                        self.env["account.move.line"].create(vals)
+        return super(AccountMove, self).post(invoice=invoice)
```
kongrattapong:
```suggestion
        return super(AccountMove, self).post()
```

kongrattapong (changes requested):

bosd:
It's an quite old PR, but can the issue with runbot be solved?

celm1990:
@pedrobaeza @max3903 can reopen this PR or I can create a new PR????

PR #293, opened 2019-11-20
[12.0] ADD account_payment_select_cost_account

eLBati:
The module allows users to specify which account to use in outbound payments.
Without this module, you can only use a payable account, not being able to use a cost account.
This is useful when you need to track payment status, also attaching files and chatting, for payments not tracked by other documents, like fiscal payments,
penalties, etc.

### DIFF
--- a/setup/account_payment_select_cost_account/setup.py
+++ b/setup/account_payment_select_cost_account/setup.py
@@ -1,0 +1,6 @@
+import setuptools
+
+setuptools.setup(
+    setup_requires=['setuptools-odoo'],
+    odoo_addon=True,
+)
--- a/account_payment_select_cost_account/__init__.py
+++ b/account_payment_select_cost_account/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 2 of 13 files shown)


mrcast (approved):
Functional test ok

setup/account_payment_select_cost_account/setup.py
```diff
@@ -0,0 +1,6 @@
+import setuptools
```
kittiu, line 1: Hi @eLBati this is not a review but my curiousity
* What is setup tools is about? And normally this will be created automatically by ocabot, or we have to do it ourselves?
* If ocabot is doing it, should it present here in the PR?

kittiu:
Functional tested, and technical reviews.
Thank you for contribution.

kittiu (approved):

setup/account_payment_select_cost_account/setup.py
eLBati: Hi @kittiu,
more info about setuptools-odoo: https://pypi.org/project/setuptools-odoo

This is normally created by ocabot, so you should not worry about adding it.

I explicitly add it when I need to deploy the module before having it released on pypi, using a `requirements.txt` line like this:
```
-e git+https://github.com/eLBati/account-payment.git@d1fc63d62970091a874fcaacaf3f0aadf805b2a9#egg=odoo12-addon-account-payment-select-cost-account&subdirectory=setup/account_payment_select_cost_account
```

PR #298, opened 2019-12-20
[11.0] Implement the payment return management for payment orders (output payment type).

Nooka10:
Currently, only the payment return from debit order (pain008) are managed. But it's also possible to receive payment return from payment order (pain001)...

With this PR, the module will be able to manage both types of pain002.


Nooka10:
This PR has been replaced by this one: https://github.com/OCA/account-payment/pull/299

PR #306, opened 2020-01-29
[13.0][MIG] account_due_list

MiquelRForgeFlow:
Standard migration of account_due_list.


AdriaGForgeFlow:
Commit message should begin with [MIG] not [IMP] right?

JoanSForgeFlow (approved):
Tested on local, LGTM :+1:

account_due_list/views/payment_view.xml
```diff
@@ -0,0 +1,131 @@
...
+
+    <!--****************************** PAYMENTS ***********************************-->
+    <record id="view_payments_tree" model="ir.ui.view">
+        <field name="name">Payments</field>
+        <field name="model">account.move.line</field>
+        <field name="field_parent">partner_id</field>
+        <field name="priority" eval="20"/>
+        <field name="arch" type="xml">
+            <tree string="Payments" decoration-muted="reconciled!=False"
+                    decoration-danger="date_maturity&lt;current_date" create="false" delete="false">
```
AdriaGForgeFlow: I would suggest to put here the following:
``` xml
decoration-danger="reconciled==False and date_maturity&lt;current_date"
```
This way, payments that are already reconciled are shown in grey rather than red

AdriaGForgeFlow (approved):
Functional review :+1: 

Code review, small comment below. Could you use this PR to update company name? (eficent --> forgeflow)

MiquelRForgeFlow:
> Commit message should begin with [MIG] not [IMP] right?

haha good

MiquelRForgeFlow:
comments attended

alexandreduf:
Hi,

I'm not sure this is the right place to put this comment, but I'll try it.

The Readme's "Usage" section says : 
> To use this module, you need to go to:
> Invoicing > Accounting > Payments and due list

**Problem**
When you run the runbot, there is no accounting menu. So it's impossible for a novice user (like me) to try this module. 
Runbot_no_accounting

Thanks for the great work!

pedrobaeza:
Enable in your user "Show Full Accounting Features".

PR #310, opened 2020-02-06
[12.0] [REF] Account payment return: invoice returned vals

qgroulard:
For better extensibility, notify the invoice that a payment return has arrived and let it do what it implies for itself.

### DIFF
--- a/account_payment_return/models/account_invoice.py
+++ b/account_payment_return/models/account_invoice.py
@@ -27,6 +27,12 @@ class AccountInvoice(models.Model):
         (self - returned_invoices).filtered('returned_payment').write(
             {'returned_payment': False})
 
+    @api.multi
+    def _payment_returned(self, return_line):
+        vals = return_line._prepare_invoice_returned_vals()
+        if vals:
+            self.write(vals)
+
     @api.one
     @api.depends('payment_move_line_ids.amount_residual')
     def _get_payment_info_JSON(self):
--- a/account_payment_return/models/payment_return.py
+++ b/account_payment_return/models/payment_return.py
@@ -212,7 +212,6 @@ class PaymentReturn(models.Model):
             raise UserError(
                 _("You must input all moves references in the payment "
                   "return."))
-        invoices = self.env['account.invoice']
         move_line_model = self.env['account.move.line']
         move = self.env['account.move'].create(
             self._prepare_return_move_vals())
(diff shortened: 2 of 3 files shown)


account_payment_return/models/account_invoice.py
```diff
@@ -27,6 +27,10 @@ def check_payment_return(self):
         (self - returned_invoices).filtered('returned_payment').write(
             {'returned_payment': False})
 
+    @api.multi
+    def _payment_returned(self, return_line):
+        self.write(return_line._prepare_invoice_returned_vals())
```
pedrobaeza: Don't write if returned vals are empty.

qgroulard: Fixed.

sbidoul (approved):

account_payment_return/models/payment_return.py
```diff
@@ -228,8 +227,10 @@ def action_confirm(self):
                 # returned_moves: debit on customer account (from invoice move)
                 returned_moves = move_line.matched_debit_ids.mapped(
                     'debit_move_id')
+                returned_moves.mapped('invoice_id')._payment_returned(
```
pedrobaeza: Can this return more than one invoice and thus give singleton error?

qgroulard: This may return more than one invoice when a payment reconciled with several invoices is returned.
`_payment_returned` is api.multi, it should not be problematic.

sbidoul (approved):
The `_payment_returned` event method is now on `account.move.line` which is more flexible.

Still LGTM. 

@pedrobaeza ok for you? This small refactoring is the basis for the new features in the other PRs.

pedrobaeza:
Seems a bit weird that you have the trigger but call a prepare method in other model. Also there seems to be now 2 `_prepare_invoice_returned_vals`.

qgroulard:
> Seems a bit weird that you have the trigger but call a prepare method in other model. Also there seems to be now 2 `_prepare_invoice_returned_vals`.

Indeed it's a bit weird but we kept the method on `payment.return` to ensure compatibility with previous versions of the code.

Moreover, some data useful for the computation of invoice returned vals are stored on `payment.return.line` and not on `payment.return`  (e.g. the return reason code). 
Thus we have added a 2nd prep method on `payment.return.line`.

sbidoul:
@pedrobaeza to me it looks clean. You can see an example usage this PR enables in https://github.com/OCA/account-payment/pull/313/commits/cf983bf0ce86e68ed5bfbc7830a90bbef0644752

pedrobaeza:
I think we should keep then both `_payment_returned` for retro-compatibility as well.

sbidoul:
> I think we should keep then both _payment_returned for retro-compatibility as well.

@pedrobaeza I don't understand: `_payment_returned` is present on move.line and account.invoice, and it is new in this PR, so there is no compatibility issue.

pedrobaeza:
Yeah, sorry, I mixed things as I was seeing only one commit diff. Please squash both commits in one and I'll merge.

qgroulard:
@pedrobaeza  Commits squashed.

PR #311, opened 2020-02-06
[12.0] [ADD] Account payment invoice last return reason

qgroulard:

Add a field 'Last Returned Payment Reason' on invoices.

### DIFF
--- a/account_payment_return/models/account_invoice.py
+++ b/account_payment_return/models/account_invoice.py
@@ -27,6 +27,12 @@ class AccountInvoice(models.Model):
         (self - returned_invoices).filtered('returned_payment').write(
             {'returned_payment': False})
 
+    @api.multi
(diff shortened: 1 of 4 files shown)


account_payment_return/models/account_invoice.py
```diff
@@ -15,6 +15,9 @@ class AccountInvoice(models.Model):
         string='Payment returned',
         help='Invoice has been included on a payment that has been returned '
              'later.')
+    last_returned_payment_reason_id = fields.Many2one(
+        comodel_name='payment.return.reason'
```
sbidoul: Don't we need a `string=` here?

qgroulard: Added.

sbejaoui (approved):

PR #312, opened 2020-02-06
[12.0] [ADD] Account payment return revoke mandate

qgroulard:

Add an option on return reason to revoke mandates.


sbejaoui:
I rebased this PR on #338 to be able to add the new field revoke_mandates to the payment return reason form view

PR #313, opened 2020-02-06
[12.0] [ADD] Payment return block move_lines

qgroulard:

Add an option on 'payment.return.reason' in order to block move_lines that were reconciled with a payment that has been returned.

Use case example:
You receive a return payment for a debit order with a certain code such that you know that trying a new debit will fail, blocking the move line prevents to retry.


sbejaoui:
I rebased this PR on #338 to be able to add the new field block_move_lines to the payment return reason form view

PR #316, opened 2020-02-10
[12.0][IMP] account_payment_return, new reason action type

qgroulard:
Add a field 'Next Action' on a payment return reason.

This field is meant to categorize payment returns by the action to take when received.
For instance, 'Next Action' could be 'Renew mandate', 'Update BIC' or 'Contact Customer'.


kittiu:
Hi @qgroulard 
I am testing this PR, but AFIK, it simply add new m2o data field in the reasons. Which i don't see how it will be useful yet (i.e., no report related to it).
May be you can explain the real use of this PR ? :)
Thanks!

kittiu:
By the way, can you help change the title of this PR to be,
[12.0][IMP] account_payment_return, new reason action type
That would make it search easily. :)

qgroulard:
> Hi @qgroulard
> I am testing this PR, but AFIK, it simply add new m2o data field in the reasons. Which i don't see how it will be useful yet (i.e., no report related to it).
> May be you can explain the real use of this PR ? :)
> Thanks!

Hi @kittiu 
Indeed this PR adds a new m2o field with no further logic.
My customer needed to be able to filter returned payments based on the next thing to do.
This helps to treat them in mass and to assign actions to employees (i.e. the guy who is responsible for renewing mandates might not be the same that the one who should call the customers).

I thought it could already benefit to others as is.

kittiu:
@qgroulard thanks for explaining. But so, I am not so sure it can benefit other.
You can just add this field on your custom module, don't you?

Let's wait for others opinion :)

sbejaoui:
I rebased this PR on #338 to be able to add the new field next_action_type_id to the payment return reason form view

enricostano:
Hi @qgroulard , does this PR allows to generate an invoice to charge the customer dunning expenses? Thanks!

qgroulard:
> Hi @qgroulard , does this PR allows to generate an invoice to charge the customer dunning expenses? Thanks!

Hello @enricostano, it does not automatically.
At least it can help mark the returns as 'Next action: generate dunning invoice" so that user can easily retrieve them and create the invoices manually.

PR #302, opened 2020-01-22
[12.0][ADD] IPPay ACH eCheck payment acquirer

dreispt:

PR #301, opened 2020-01-22
[ADD] IPPay Credit Card payment acquirer

dreispt:

PR #318, opened 2020-02-24
[12.0] IPPAY: Fix Portal Manage Payments

dreispt:
Fixes error when accessing the "Manage payment methods", in the Customer Portal home page (/my/home).

With this change, the saved payment tokens list is for the logged in user, so testes must be done with the customer's user logged in to the Portal.

(Depends on #301 and #302, and previous commits are coming from there)

Note that the saved methods list is not grouped nor ordered by payment method provider: 
we first see the "add payment" forms for each acquirer, and then see the list of saved tokens.

This is true both for the "Manage Payment methods" page and the "Pay Now" dialog box.

image

image


dreispt:
Superseded by #319 and #320

PR #319, opened 2020-03-11
[12.0][ADD] IPPay Payment Acquirer for Credit Cards, v1.1.0

dreispt:
This version is compatible with the "Manage Payment Methods" option in the Customer Portal home page.


dreispt:
Added a fix found during tests.
There are three ways you can add a credit card: on the Customer form, on the Invoice Payment portal page, an on the My Payment Methods portal page. This issue was found on the later one.

PR #320, opened 2020-03-11
[12.0][ADD] IPPay Payment Acquirer for ACH eChecks, v1.1.0

dreispt:
This version is compatible with the "Manage Payment Methods" option in the Customer Portal home page.


dreispt:
This is still valid, and deployed in production.

PR #324, opened 2020-03-25
new module payment_bitcoin

jans23:
This module lets ecommerce customers pay with Bitcoin.


bosd:
Would be great to have this module!

bosd:
Thanks for all this work!!
Have you considered doing a full integration with a Bitcoin RPC?
Looks a bit cumbersome, to have the user pre-gerenate bitcoin adresses.

In the readme there is a TODO for the currency exchange rate.
FYI, there is an Coin Market Cap exchange rate module in the OCA currency repo.

p.s. Please change title to [12.0[ADD]

jans23:
User pre-gerenate Bitcoin adresses were easier to implement and allow to keep the wallet separate (potentially offline) from Odoo. This is why we will stick to that approach.

Using currency_rate_update_cmc for the currency exchange would indeed be an option. But the exchange rate varies between different exchange platforms and we would need to check if Coindesk is suitable for us.

This PR will soon be superseeded by a PR for 15.0 but essentially the same module, same features.

bosd:
> User pre-gerenate Bitcoin adresses were easier to implement and allow to keep the wallet separate (potentially offline) from Odoo. This is why we will stick to that approach.

Good point to keep the wallet separated.
Yet, i'll stick to my argument that from usability it is desired to use an wallet rpc.
An Wallet RPC can also be used offline to generate the adresses.
Hopefully you consider it as an option.
You can propably re-use a lot of code from the Monero integration I'm working on.
https://github.com/bosd/moneroodoo


Did only a quick browse in this pr code.
I did'nt see where the Bitcoin currency is defined.

Have you considered the account_cryptocurrency as a dependency?
How is this one integrated into the accounting?



> This PR will soon be superseeded by a PR for 15.0 but essentially the same module, same features.

Would love to see that one coming.

jans23:
A wallet RPC would be welcome but that would be a new version, not something we will implement now.

In our use case we accounting/legally don't use Bitcoins directly because we involve a currency exchange which only transfers the Euro to us. We don't really use Bitcoin currency other than in communicating the payment details with the customer.

PR #333, opened 2020-06-02
[10.0][ADD] account_followup_email_template

pepetreshere:
Hello, we would like to have this module considered into the OCA repositories, thank you very much. It allows users to more easily send follow up emails by including a pre-built email template


pepetreshere:
Hello @pedrobaeza, would you please consider the inclusion of this feature in OCA repositories?, or perhaps provide me some guidance on who should I contact instead?. Of course if you also have some technical suggestions I would like to know.

pedrobaeza:
Well, if there are reviewers interested in it, it can be included, but it doesn't depend on me.

There are some similar modules though, being the main one: https://github.com/OCA/credit-control/tree/12.0/account_credit_control, as full credit followup.

PR #303, opened 2020-01-23
WIP: 12.0 mig account payment batch process

ovnicraft:

PR #337, opened 2020-06-30
[12.0][MIG] account_payment_batch_process

murtuzasaleh:
Batch Payments Processing
======================
This module allows us to process payments in batch for customer's invoices and vendor bills.


astirpe:
There is already https://github.com/OCA/account-payment/pull/303

PR #343, opened 2020-07-29
[13.0][MIG] account_voucher_killer

astirpe:


astirpe:
Tests of `account_payment_return` are failing due to odoo/odoo#55088

PR #347, opened 2020-09-14
Adding path to activate Full Accounting Settings

alexandreduf:
Since the Runbot doesn't have Full Accounting Settings already activated, I think it's better to explain how to get it.

### DIFF
--- a/account_due_list/readme/USAGE.rst
+++ b/account_due_list/readme/USAGE.rst
@@ -1,3 +1,3 @@
-To use this module, you need to go to:
-
-* Invoicing > Accounting > Payments and due list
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I would put that in the CONFIGURE.rst file, as it's a configuration thing.

alexandreduf:
Thanks. Just created CONFIGURE.rst.

pedrobaeza:
I still see the old one.

alexandreduf:
The old what?

pedrobaeza:
The old change in USAGE.rst instead of CONFIGURE.rst.

PR #348, opened 2020-09-14
Adding path to activate Full Accounting Settings

alexandreduf:
Since the Runbot doesn't have Full Accounting Settings already activated, I think it's better to explain how to get it.


pedrobaeza (approved):
We usually use the format of numbered list for each step (`#.`), but I'm not pushing you for more changes. Take into account for next time (the same as the commit message as pointed in other PR).

alexandreduf:
Thank you! I'll try to do that next time!

marcelsavegnago (approved):

PR #329, opened 2020-05-18
[13.0][MIG] account_check_printing_report_base

Saran440:
depend on report_wkhtmltopdf_param https://github.com/OCA/reporting-engine/pull/402

- Delete model account.register.payment. In odoo13 used account.payment instead.
- Fix Test Script for support invoice v13

PR #351, opened 2020-09-18
[13.0] [MIG] account_check_printing_report_base: Migration to v13.0

CarlosRoca13:
cc @Tecnativa TT25537

Please @carlosdauden @sergio-teruel, can you review this?


account_check_printing_report_base/README.rst
```diff
@@ -0,0 +1,109 @@
...
+* Ursa Information Systems
+
+Contributors
+~~~~~~~~~~~~
+
+* Jordi Ballester Alomar <jordi.ballester@eficent.com>
+* Lois Rilo Antelo <lois.rilo@eficent.com>
+* Sandip Mangukiya <smangukiya@ursainfosystems.com>
+* Luis M. Ontalba <luis.martinez@tecnativa.com>
+* Manuel Marquez <mmarquez@iterativo.do>
```
joao-p-marques, line 84: Seems like you didn't regenerate the README, and the changes to the Contributors file didn't apply.

account_check_printing_report_base/README.rst
```diff
@@ -0,0 +1,109 @@
...
+**Table of contents**
+
+.. contents::
+   :local:
+
+Installation
+============
+
+In order to install this module you must first install also the module
+'report_paper_wkhtmltopdf_params', available in
```
joao-p-marques: I'm not sure, but if you need to install this module first, shouldn't it be added as a dependency in the manifest?

account_check_printing_report_base/tests/test_check_printing_report.py
```diff
@@ -0,0 +1,157 @@
...
+        self.account_account_model = self.env["account.account"]
+        self.payment_model = self.env["account.payment"]
+
+        self.partner1 = self.env.ref("base.res_partner_1")
+        self.company = self.env.ref("base.main_company")
+        self.currency_usd_id = self.env.ref("base.USD").id
+        self.currency_euro_id = self.env.ref("base.EUR").id
+        self.acc_payable = self.env.ref("account.data_account_type_payable")
+        self.acc_expense = self.env.ref("account.data_account_type_expenses")
+        self.product = self.env.ref("product.product_product_4")
```
joao-p-marques, line 30: You should create test data instead of using demo data. That way the test will be more resilient and independent from previous operations on the DB

CarlosRoca13: In this way I use the test created on the previous version, I usually make what you say. 

Probably the best way is change the declaration to avoid what you say, but doing like is done we re-use the code from previous versions and I think that this solution could be repetitive, because I will define at that recorsets the same data that is declared yet on demo data. What do you think?

joao-p-marques: Well, there is no problem in creating new data just for the test, even if similar records already exist as demo. By doing this, you make sure your tests won't fail just because a test from a previous model changed or deleted the record you are trying to use. 

However, if the code already comes from the previous version and is working, maybe somebody else could give their opinion on if it is necessity to re-write. @carlosdauden @sergio-teruel

pedrobaeza:
This must include the changes we are doing now in v12

CarlosRoca13:
@pedrobaeza Changes added

account_check_printing_report_base/tests/test_check_printing_report.py
```diff
@@ -138,15 +132,11 @@ def test_check_printing_with_layout(self):
...
-            "account payable test", "ACPRB1", self.acc_payable, True
-        )
-        vendor_bill = self._create_vendor_bill(acc_payable)
         acc_expense = self._create_account(
             "account expense test", "ACPRB2", self.acc_expense, False
         )
-        self._create_invoice_line(acc_expense, vendor_bill)
-        vendor_bill.action_invoice_open()
+        vendor_bill = self._create_vendor_bill(acc_expense)
+        vendor_bill.state = "posted"
```
ernestotejeda: I think you should use action_post method instead
```suggestion
        vendor_bill.action_post()
```

CarlosRoca13:
Change done! @ernestotejeda

ernestotejeda:
@CarlosRoca13 ,runbot is red, check it out please.

pedrobaeza:
I don't see improvement from 12.0 in a separate commit. It seems instead that it has been merged in the 13.0 migration. Please cherry-pick it as is before 13.0 migration, pass black phase again and commit in their corresponding, and then perform 13.0 migration. You have to check also for a warning that appears and make runbot red.

CarlosRoca13:
@pedrobaeza Changes done!

pedrobaeza:
@CarlosRoca13 there's another PR migrating this to version 13 that is prior to yours: #329

account_check_printing_report_base/readme/USAGE.rst
```diff
@@ -0,0 +1,4 @@
+* Go to 'Invoicing / Purchases / Payments'. Select one of the payments with
```
Saran440:
```suggestion
* Go to 'Invoicing / Vendors / Payments'. Select one of the payments with
```

Saran440 (approved):
Functional Test :+1:  and Code reviewed.

Could you fix test script to cover case? see example from https://github.com/OCA/account-payment/blob/18c2db7e61707854db761cf4df9253634e1b5fd7/account_check_printing_report_base/tests/test_check_printing_report.py

CarlosRoca13:
@Saran440 Thanks!, changes done!

joao-p-marques (approved):

PR #357, opened 2020-10-03
[13.0][IMP] account_payment_multi_deduction, add hook when create deduction move line

kittiu:
Just make sub function, so I can hook it in other addon. Can we fast track this?

cc @pedrobaeza @newtratip


pedrobaeza (approved):
Although not using this module, the change seems harmless and help the extensions

kittiu:
Thank you!

PR #359, opened 2020-10-03
[13.0][IMP] account_payment_multi_deduction, ability to multi deduct with keep open

kittiu:
Currently there is no way to register payment with multiple deduction at the same time keep some amount open.

This PR, user can flag some line in deduction table, to keep some amount open. And that linke, i..e, 100.0 will be reduced from AR/AP account, in order to keep it open.

image


kittiu:
Test added, ready for review.

PR #362, opened 2020-10-23
[12.0] [FIX+IMP] account_payment_promissory_note: Fixed a bug with invoce multi payment and add code to allow print promisory notes

CarlosRoca13:
cc @Tecnativa TT25536

@sergio-teruel @pedrobaeza this are the changes for avoid reapeted code and solve the bug that I found on the v13 migration.

### DIFF
--- a/account_check_printing_report_base/data/report_paperformat.xml
+++ b/account_check_printing_report_base/data/report_paperformat.xml
@@ -15,4 +15,20 @@
         <field name="header_spacing">3</field>
         <field name="dpi">80</field>
     </record>
+
+    <record id="paperformat_euro_no_margin" model="report.paperformat">
+        <field name="name">Promissory Note Letter</field>
+        <field name="default" eval="False" />
+        <field name="format">custom</field>
+        <field name="page_height">297</field>
+        <field name="page_width">210</field>
+        <field name="orientation">Portrait</field>
+        <field name="margin_top">40</field>
+        <field name="margin_bottom">0</field>
+        <field name="margin_left">0</field>
+        <field name="margin_right">0</field>
+        <field name="header_line" eval="False" />
+        <field name="header_spacing">40</field>
+        <field name="dpi">90</field>
+    </record>
 </odoo>
--- a/account_check_printing_report_base/report/__init__.py
+++ b/account_check_printing_report_base/report/__init__.py
@@ -1,1 +1,2 @@
 from . import check_print
+from . import lang
--- a/account_check_printing_report_base/report/check_print.py
+++ b/account_check_printing_report_base/report/check_print.py
@@ -7,6 +7,7 @@
 import time
 from odoo import api, models
 from odoo.tools import float_is_zero
+from . import lang
 
 
 class ReportCheckPrint(models.AbstractModel):
@@ -118,3 +119,28 @@ class ReportCheckPrint(models.AbstractModel):
 class ReportCheckPrintA4(models.AbstractModel):
     _name = 'report.account_check_printing_report_base.report_check_base_a4'
     _inherit = 'report.account_check_printing_report_base.report_check_base'
+
+
+class ReportPromissoryNotePrint(models.AbstractModel):
+    _name = 'report.account_check_printing_report_base.promissory_footer_a4'
+    _inherit = 'report.account_check_printing_report_base.report_check_base_a4'
+
+    def fill_stars(self, amount_in_word):
+        if amount_in_word and len(amount_in_word) < 100:
+            stars = 100 - len(amount_in_word)
+            return ' '.join([amount_in_word, '* ' * stars])
+        else:
+            return amount_in_word
+
+    def amount2words(self, amount):
+        return lang.num2words_custom(
+            amount, to='currency', lang=self.env.context.get('lang', 'en_US'))
+
+    @api.model
+    def _get_report_values(self, docids, data=None):
+        res = super()._get_report_values(docids, data=data)
+        res.update({
+            'fill_stars': self.fill_stars,
+            'num2words': self.amount2words,
+        })
+        return res
--- a/account_payment_promissory_note/wizard/account_register_payments.py
+++ b/account_payment_promissory_note/wizard/account_register_payments.py
@@ -6,30 +6,16 @@ from odoo import api, models
 
 
 class AccountRegisterPayments(models.TransientModel):
-    _name = 'account.register.payments'
-    _inherit = ['account.register.payments', 'account.abstract.payment']
+    _inherit = 'account.register.payments'
 
-    def get_payment_vals(self):
-        vals = super(AccountRegisterPayments, self).get_payment_vals()
-        vals.update({
-            'promissory_note': self.promissory_note,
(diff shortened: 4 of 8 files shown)


CarlosRoca13:
@pedrobaeza Done changes talked

account_check_printing_report_base/data/report_paperformat.xml
```diff
@@ -15,4 +15,20 @@
...
+        <field name="page_width">210</field>
+        <field name="orientation">Portrait</field>
+        <field name="margin_top">40</field>
+        <field name="margin_bottom">0</field>
+        <field name="margin_left">0</field>
+        <field name="margin_right">0</field>
+        <field name="header_line" eval="False" />
+        <field name="header_spacing">40</field>
+        <field name="dpi">90</field>
+    </record>
```
victoralmau, line 33: It is created but never used, is correct?

account_check_printing_report_base/report/lang_ES.py
```diff
@@ -0,0 +1,34 @@
...
+    if lang != 'es':
+        return num2words(number, ordinal, lang, to, **kwargs)
+    else:
+        # Fixed implementation
+        converter = Num2WordESCustom()
+        if ordinal:
+            return converter.to_ordinal(number)
+        if to not in CONVERTES_TYPES:
+            raise NotImplementedError()
+        return getattr(converter, 'to_{}'.format(to))(number, **kwargs)
```
victoralmau:
```suggestion
# pylint: disable=missing-import-error
from num2words import num2words
from num2words import CONVERTER_CLASSES, CONVERTES_TYPES
from num2words.lang_ES import Num2Word_ES

num2words_by_lang = {
    'es': 'Num2WordESCustom'
}

class Num2WordESCustom(Num2Word_ES):
    # Replace centavo to céntimo. Original method copy
    def to_currency(self, val, longval=True, old=False):
        hightxt, lowtxt = "euro/s", u"céntimo/s"
        if old:
            hightxt, lowtxt = "peso/s", "peseta/s"
        result = self.to_splitnum(val, hightxt=hightxt, lowtxt=lowtxt,
                                  divisor=1, jointxt="y", longval=longval)
        # Handle exception, in spanish is "un euro" and not "uno euro"
        return result.replace("uno", "un")


def num2words(number, ordinal=False, lang='en', to='cardinal', **kwargs):
    if lang not in CONVERTER_CLASSES:
        # ... and then try only the first 2 letters
        lang = lang[:2]
    if lang not in CONVERTER_CLASSES:
        raise NotImplementedError()

    if lang not in num2words_by_lang:
        return num2words(number, ordinal, lang, to, **kwargs)
    else:
        # Fixed implementation
        converter = num2words_by_lang[lang]()
        if ordinal:
            return converter.to_ordinal(number)
        if to not in CONVERTES_TYPES:
            raise NotImplementedError()
        return getattr(converter, 'to_{}'.format(to))(number, **kwargs)
```

Although work fine IMO need to standard function to add more langs in the future.

account_check_printing_report_base/report/__init__.py
```diff
@@ -1 +1,2 @@
 from . import check_print
+from . import lang_ES
```
victoralmau:
```suggestion
from . import lang
```

account_check_printing_report_base/report/check_print.py
```diff
@@ -7,6 +7,7 @@
 import time
 from odoo import api, models
 from odoo.tools import float_is_zero
+from . import lang_ES
```
victoralmau:
```suggestion
from . import lang
```

account_check_printing_report_base/report/check_print.py
```diff
@@ -115,6 +116,31 @@ def _get_report_values(self, docids, data=None):
...
+
+    def fill_stars(self, amount_in_word):
+        if amount_in_word and len(amount_in_word) < 100:
+            stars = 100 - len(amount_in_word)
+            return ' '.join([amount_in_word, '* ' * stars])
+        else:
+            return amount_in_word
+
+    def amount2words(self, amount):
+        return lang_ES.num2words_es(
```
victoralmau:
```suggestion
        return lang.num2words(
```

account_payment_promissory_note/wizard/account_register_payments.py
```diff
@@ -6,31 +6,17 @@
...
-        action = self.env.ref(
-            'account.action_account_payments_payable').read()[0]
-        action['views'] = [
-            (self.env.ref('account.view_account_payment_form').id, 'form')]
-        action['res_id'] = payment.id
-        return action
-
     @api.onchange('promissory_note')
     def _onchange_promissory_note(self):
         super()._onchange_promissory_note()
```
victoralmau, line 22: Why it's necessary?

victoralmau (changes requested):

account_check_printing_report_base/report/lang_ES.py
CarlosRoca13: The other langs are considerated on Odoo Core num2words implementation, the problem is that the spanish lang is not correct because they traduce "cents" as "centavos" and this should be "céntimos". That's why we use lang_ES name

account_payment_promissory_note/wizard/account_register_payments.py
CarlosRoca13: Because we are overwriting the method of `account.abstract.payment` for compute the invoices on `account.register.payments`

account_check_printing_report_base/data/report_paperformat.xml
CarlosRoca13: Yes, it is correct, but it will be used on other 3 modules and it is added here for avoid repeated code.

account_payment_promissory_note/wizard/account_register_payments.py
victoralmau: But `account.abstract.payment` is not `_inherit` now in wizard.

account_check_printing_report_base/report/lang_ES.py
victoralmau: Ok, i know about it and all this code is needed but i only suggested rename file and functions to add more function langs in future without change anything.

account_payment_promissory_note/wizard/account_register_payments.py
CarlosRoca13: Because in the main definition is inherited https://github.com/odoo/odoo/blob/b09660c6407f75024d23d2eb785c7ce83f0549ad/addons/account/models/account_payment.py#L281

CarlosRoca13:
@victoralmau Change suggested done, with a little change, I set to the custom num2words method name to num2words_custom because we don't want to lose the num2words main mathod

victoralmau:
> @victoralmau Change suggested done, with a little change, I set to the custom num2words method name to num2words_custom because we don't want to lose the num2words main mathod

Ok perfect.

victoralmau (approved):

pedrobaeza (approved):

PR #366, opened 2020-10-29
[13.0] [IMP] account_check_printing_report_base: Added IMP set on v12

CarlosRoca13:
cc @Tecnativa TT25536

Added the improvement that set the fix of spanish lang, and code to avoid reapeat it on specific bank modules.

Please @pedrobaeza @victoralmau can you review this?


victoralmau:
Have do you think about add some little tests related to this new report?

victoralmau (approved):

pedrobaeza (approved):

PR #367, opened 2020-10-29
[13.0] [MIG] account_payment_promissory_note: Migration to v13

CarlosRoca13:
cc @Tecnativa TT25536

Please review this @carlosdauden @victoralmau


victoralmau (approved):

account_payment_promissory_note/wizard/account_register_payments.py
```diff
@@ -21,8 +22,8 @@ def _onchange_promissory_note(self):
         super()._onchange_promissory_note()
         if not self.date_due and self.promissory_note:
             invoices = False
-            if self._name == "account.register.payments":
+            if self._name == "account.payment.register":
```
pedrobaeza: This is not needed (even in v12, I missed this one...). The onchange is already in the proper model this `if` is going to be always True.

pedrobaeza (changes requested):

CarlosRoca13:
@pedrobaeza change done

pedrobaeza (approved):

PR #372, opened 2020-11-18
[13.0][IMP] account_payment_promissory_note: Changes on due_date

CarlosRoca13:
Applied changes on due_date for take care about the partner groups on register payments.

On one hand, if the field due_date is empty the camp will take the latest due_date of the invoices that form the payment.

On the other hand, if we set a value on due_date field, all the payments generated will take this value.

On next gif you can see how it is working.

Please @carlosdauden @sergio-teruel, can you review this?

### DIFF
--- a/account_payment_promissory_note/models/account_payment.py
+++ b/account_payment_promissory_note/models/account_payment.py
@@ -25,5 +25,6 @@ class AccountPayment(models.Model):
         super()._onchange_promissory_note()
         if not self.date_due and self.promissory_note:
             invoices = self.invoice_ids
-            if invoices:
+            same_partner = len(invoices.mapped("partner_id")) == 1
+            if invoices and same_partner:
                 self.date_due = max(invoices.mapped("invoice_date_due"))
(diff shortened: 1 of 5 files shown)


account_payment_promissory_note/models/account_payment.py
```diff
@@ -25,5 +25,10 @@ def _onchange_promissory_note(self):
         super()._onchange_promissory_note()
         if not self.date_due and self.promissory_note:
             invoices = self.invoice_ids
-            if invoices:
+            partner = invoices[0].partner_id
+            same_partner = True
+            for invoice in invoices:
+                if invoice.partner_id != partner:
+                    same_partner = False
```
pedrobaeza:
```suggestion
            same_partner = len(invoices.mapped("partner_id")) == 1
```

account_payment_promissory_note/wizard/account_register_payments.py
```diff
@@ -23,5 +30,10 @@ def _onchange_promissory_note(self):
         if not self.date_due and self.promissory_note:
             active_ids = self._context.get("active_ids")
             invoices = self.env["account.move"].browse(active_ids)
-            if invoices:
+            partner = invoices[0].partner_id
```
pedrobaeza: Same as previous

CarlosRoca13:
@pedrobaeza changes done

pedrobaeza (approved):

victoralmau (approved):

PR #376, opened 2020-11-27
FIX: use field supplier_rank instead of supplier

celm1990:
field supplier not exits any more on V13


pedrobaeza:
I would directly remove the domain for allowing any partner.

pedrobaeza (approved):

pedrobaeza:
Fast-tracking this as straight-forward:

/ocabot merge patch

PR #377, opened 2020-12-04
[12.0] [IMP] account_payment_promissory_note: Changes on due_date

CarlosRoca13:
cc @Tecnativa TT27088

Applied changes on due_date for take care about the partner groups on register payments.

On one hand, if the field due_date is empty the camp will take the latest due_date of the invoices that form the payment.

On the other hand, if we set a value on due_date field, all the payments generated will take this value.

On next gif you can see how it is working.

Please @sergio-teruel @carlosdauden review this


pedrobaeza (approved):

cubells (approved):
:+1: 

LGTM

Code review and test

carlosdauden (approved):

PR #382, opened 2021-01-07
[13.0][IMP] account_payment_term_extension: Add amount_untaxed to value option in payment term lines

victoralmau:
Add `amount_untaxed` to value option in payment term lines (%) related to invoice.

Please, @joao-p-marques  review it.
@Tecnativa TT27625

### DIFF
--- a/account_payment_term_extension/models/account_move.py
+++ b/account_payment_term_extension/models/account_move.py
@@ -1,0 +1,14 @@
+# Copyright 2021 Tecnativa Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from odoo import models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def _recompute_payment_terms_lines(self):
+        super(
+            AccountMove, self.with_context(last_stock_move=self)
+        )._recompute_payment_terms_lines()
--- a/account_payment_term_extension/models/account_payment_term.py
+++ b/account_payment_term_extension/models/account_payment_term.py
@@ -3,6 +3,7 @@
 # Copyright 2015-2016 Akretion
 # (Alexis de Lattre <alexis.delattre@akretion.com>)
 # Copyright 2018 Simone Rubino - Agile Business Group
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 import calendar
@@ -76,6 +77,24 @@ class AccountPaymentTermLine(models.Model):
     )
     months = fields.Integer(string="Number of Months")
     weeks = fields.Integer(string="Number of Weeks")
+    value = fields.Selection(
+        selection_add=[("amount_untaxed", "Percent (Untaxed amount)")]
+    )
+    value_amount_untaxed = fields.Float(string="Untaxed amount (Percent)")
+
+    @api.constrains("value", "value_amount_untaxed")
+    def _check_value_amount_untaxed(self):
+        for term_line in self:
+            if (
+                term_line.value == "amount_untaxed"
+                and not 0 <= term_line.value_amount_untaxed <= 100
+            ):
+                raise ValidationError(
+                    _(
+                        "Percentages on the Payment Terms lines "
+                        "must be between 0 and 100."
+                    )
+                )
 
     def compute_line_amount(self, total_amount, remaining_amount, precision_digits):
         """Compute the amount for a payment term line.
@@ -92,6 +111,11 @@ class AccountPaymentTermLine(models.Model):
             return float_round(self.value_amount, precision_digits=precision_digits)
         elif self.value == "percent":
             amt = total_amount * (self.value_amount / 100.0)
+            if self.amount_round:
+                amt = float_round(amt, precision_rounding=self.amount_round)
+            return float_round(amt, precision_digits=precision_digits)
+        elif self.value == "amount_untaxed":
+            amt = total_amount * self.value_amount_untaxed / 100.0
             if self.amount_round:
                 amt = float_round(amt, precision_rounding=self.amount_round)
             return float_round(amt, precision_digits=precision_digits)
@@ -173,10 +197,10 @@ class AccountPaymentTerm(models.Model):
         return date
 
     def compute(self, value, date_ref=False, currency=None):
-        """Complete overwrite of compute method to add rounding on line
-        computing and also to handle weeks and months
-        """
+        """Complete overwrite of compute method for adding extra options."""
+        # FIXME: Find an inheritable way of doing this
         self.ensure_one()
+        last_stock_move = self.env.context.get("last_stock_move", False)
         date_ref = date_ref or fields.Date.today()
         amount = value
         result = []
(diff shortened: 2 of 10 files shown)


account_payment_term_extension/models/account_move.py
```diff
@@ -0,0 +1,16 @@
...
+
+from odoo import models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def _recompute_payment_terms_lines(self):
+        context_prev = dict(self.env.context)
+        self.env.context = context_prev
```
joao-p-marques: Why is this assignment necessary?

account_payment_term_extension/models/account_payment_term.py
```diff
@@ -173,10 +190,10 @@ def apply_payment_days(self, line, date):
         return date
 
     def compute(self, value, date_ref=False, currency=None):
-        """Complete overwrite of compute method to add rounding on line
-        computing and also to handle weeks and months
-        """
         self.ensure_one()
+        last_stock_move = False
+        if "last_stock_move" in self.env.context:
+            last_stock_move = self.env.context.get("last_stock_move")
```
joao-p-marques: This could be simplified:
```suggestion
        last_stock_move = self.env.context.get("last_stock_move", False)
```

joao-p-marques (commented):
Code review and looks good :+1: 

Just some points:
Would this require a version bump in the manifest?
I think the commit prefix that applies is rather [IMP], not [ADD]. That can be confusing.
Also, check the comments bellow.

account_payment_term_extension/models/account_move.py
victoralmau: Really, odoo use `_compute_payment_terms` (https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_move.py#L880) to define amount and after use `compute` function (https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_payment_term.py#L34) in payment term.
IMO, use `amount_untaxed` from invoice need to add invoice to context to use it later because other possibilities change code so much and can't reused to other addons.

victoralmau:
> Would this require a version bump in the manifest?

I don't understand these comment.

account_payment_term_extension/models/account_move.py
joao-p-marques: Yes, I undertand that. What I meant was that, in line 13, you are assigning `context_prev` to `self.env.context`, but `context_prev` was already `self.env.context`, right? That seems unnecessary... Or am I missing something?

joao-p-marques:
I mean that, if you are introducing a new functionality to the module, 
wouldn't that require a version increase in the module's manifest?

account_payment_term_extension/models/account_move.py
victoralmau: I use that to leave the context as it was later.

victoralmau:
I've never changed the manifest version in a IMP feature (not sure it's necessary).

pedrobaeza:
The version bump is done when merging and in this case decided by the merger. The only manual version bump is done when adding migration scripts, for being sure the version of such scripts matches at least the lower one in the manifest.

account_payment_term_extension/models/account_move.py
pedrobaeza: This is not correct. All contexts changes must be done through `with_context`, but I don't think you need some context handling for doing this, but instead overwriting the proper method.

joao-p-marques:
Ok, thank you for the explanation :+1:

account_payment_term_extension/models/account_move.py
victoralmau: I have changed it to `with_context` now. I have thought about some possible solutions but IMO it's the best way to improve without change everything. Can you propose other better solution?

pedrobaeza: AFAIK, whole `compute` method was overridden in this module, so you just need to add the `if` part in that method.

victoralmau: Currently `compute_line_amount` is already used but to that function we must send an amount that we do not have (`amount_untaxed` of the invoice).

The `compute` function in `account.payment.term` model is very useful but it gets amonunt of the total of the invoice to calculate all the lines and this is insufficient since it is necessary to keep that value for the balance calculation but the amount_untaxed of the invoice is also needed to use it in some line. For this reason `with_contect()` is used.

account_payment_term_extension/views/account_payment_term.xml
```diff
@@ -37,7 +37,14 @@
         <field name="name">account.payment.term.line.form.extension</field>
         <field name="model">account.payment.term.line</field>
         <field name="inherit_id" ref="account.view_payment_term_line_form" />
+        <field name="priority">99</field>
         <field name="arch" type="xml">
+            <xpath expr="//form/group/group[2]/div/span" position="replace">
```
pedrobaeza: Don't use `replace`, but attributes, and use a better selector, as this one is very fragile.

account_payment_term_extension/readme/DESCRIPTION.rst
```diff
@@ -1,5 +1,6 @@
 This module extends the functionality of payment terms to :
 
+* select amount_untaxed type in lines (related to amount_untaxed from invoice)
```
pedrobaeza: This is functional documentation, not technical one

```suggestion
* select "Percent (untaxed amount)" type in lines for using the base amount instead of the total (with taxes) one.
```

account_payment_term_extension/models/account_payment_term.py
```diff
@@ -187,7 +202,15 @@ def compute(self, value, date_ref=False, currency=None):
         precision_digits = currency.decimal_places
         next_date = fields.Date.from_string(date_ref)
         for line in self.line_ids:
-            amt = line.compute_line_amount(value, amount, precision_digits)
+            if line.value == "amount_untaxed" and last_stock_move:
+                amount2 = amount
+                if amount2 == value:
```
pedrobaeza: Why do you need this comparison?

account_payment_term_extension/models/account_payment_term.py
```diff
@@ -76,6 +77,22 @@ class AccountPaymentTermLine(models.Model):
...
+    )
+
+    @api.constrains("value", "value_amount")
+    def _check_percent(self):
+        for term_line in self:
+            if term_line.value == "amount_untaxed" and (
+                term_line.value_amount < 0.0 or term_line.value_amount > 100.0
+            ):
+                raise ValidationError(
+                    _(
```
pedrobaeza, line 93: Add test for this

account_payment_term_extension/views/account_payment_term.xml
pedrobaeza: You can avoid this not reusing same percentage field. I would recommend this instead.

account_payment_term_extension/models/account_payment_term.py
```diff
@@ -76,6 +77,22 @@ class AccountPaymentTermLine(models.Model):
     )
     months = fields.Integer(string="Number of Months")
     weeks = fields.Integer(string="Number of Weeks")
+    value = fields.Selection(
+        selection_add=[("amount_untaxed", "Amount untaxed (percent)")]
```
pedrobaeza:
```suggestion
        selection_add=[("amount_untaxed", "Percent (Untaxed amount)")]
```

account_payment_term_extension/models/account_payment_term.py
```diff
@@ -76,6 +77,22 @@ class AccountPaymentTermLine(models.Model):
...
+    )
+
+    @api.constrains("value", "value_amount")
```
pedrobaeza: Don't reuse the same field for this percentage.

pedrobaeza (changes requested):

account_payment_term_extension/models/account_payment_term.py
victoralmau: Removed now. Really it's not necessary according to tests.

account_payment_term_extension/views/account_payment_term.xml
victoralmau: Removed now.

pedrobaeza (changes requested):
I don't get your changes, as now the percentage field is not editable. What I told you was:

- Create a new field for storing the percentage of the base amount and don't reuse the existing one.
- The constraint will be over that new field.
- Show that field in the form when applicable.

pedrobaeza (approved):

PR #383, opened 2021-01-08
[13.0][IMP] account_payment_term_extension: Make tests more resilient

joao-p-marques:
A `account.journal` record was being searched and used to create invoices, even though it was not necessary.
This introduced the problem of finding more than one record that satisfied the search query, breaking the test for a singleton error.
This removes the need for that.

@Tecnativa
TT27288

ping @pedrobaeza

### DIFF
--- a/account_payment_term_extension/tests/test_account_payment_term_multi_day.py
+++ b/account_payment_term_extension/tests/test_account_payment_term_multi_day.py
@@ -9,11 +9,8 @@ class TestAccountPaymentTermMultiDay(common.TransactionCase):
         super(TestAccountPaymentTermMultiDay, self).setUp()
         self.payment_term_model = self.env["account.payment.term"]
         self.invoice_model = self.env["account.move"]
-        journal_model = self.env["account.journal"]
-        self.journal = journal_model.search([("type", "=", "purchase")])
-        self.partner = self.env.ref("base.res_partner_3")
-        self.product = self.env.ref("product.product_product_5")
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

CarlosRoca13 (approved):

chienandalu (approved):

pedrobaeza:
Fast-tracking for solving integration tests:

/ocabot merge patch

pedrobaeza:
This is not FIX but IMP, and it's better to switch to `Form` for avoiding future problems as you are doing the changes.

joao-p-marques:
Ok, but I ended up not switching to `Form`. Rather, I just removed the reference to `journal_id` from the `create` method call.

pedrobaeza:
Yes, but as said, this will blow up in the future if not switching. And the commit title should be `[IMP] account_payment_term_extension: Make tests more resilient`

joao-p-marques:
Thank you! I just pushed the change with the new commit message.

pedrobaeza:
I still don't see the use of `Form`

joao-p-marques:
Sorry, I misunderstood your previous message. Making the changes now.

joao-p-marques:
@pedrobaeza the latest version uses `Form` for the creation of all invoices. 
Also, I ended up using the `line_form.tax_ids.clear()` approach, as oppose to defining the product without taxes in the 1st place, as I couldn't find how to do it. I can have a closer look at it tomorrow, though.

PR #387, opened 2021-01-21
[12.0][FIX] account_check_printing_report_base: Tens in word cents are showed as units (Only ES)

carlosdauden:
Improve code using evolutioned base method instead of custom method.

Avoid overwrite precision…(4.90 returns four dot nine instead of four dot ninety)
https://github.com/savoirfairelinux/num2words/commit/1c699d1bb411f42b37f619fb6560e6f03d67c463#diff-28b986a636d0cd36a886fb49a7dcde88885ec58befc6dbb7901b3ffe54a35a20R112

Before this PR:
4.90 returns "cuatro euros con nueve céntimos"

Before this PR:
4.90 returns "cuatro euros con noventa céntimos"

@Tecnativa TT27765

### DIFF
--- a/account_check_printing_report_base/report/lang.py
+++ b/account_check_printing_report_base/report/lang.py
@@ -2,6 +2,7 @@
 from num2words import num2words
 from num2words import CONVERTER_CLASSES, CONVERTES_TYPES
 from num2words.lang_ES import Num2Word_ES
+from num2words.base import Num2Word_Base
 
 num2words_by_lang = {
     'es': 'Num2WordESCustom'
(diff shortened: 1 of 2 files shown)


CarlosRoca13 (approved):

account_check_printing_report_base/report/lang.py
```diff
@@ -2,22 +2,22 @@
...
+    # TODO: PR to remove overwrite in num2words code and use CURRENCY_FORMS
     def to_currency(self, val, longval=True, old=False):
-        hightxt, lowtxt = "euro/s", u"céntimo/s"
-        if old:
-            hightxt, lowtxt = "peso/s", "peseta/s"
-        result = self.to_splitnum(val, hightxt=hightxt, lowtxt=lowtxt,
-                                  divisor=1, jointxt="y", longval=longval)
-        # Handle exception, in spanish is "un euro" and not "uno euro"
-        return result.replace("uno", "un")
+        return Num2Word_Base.to_currency(self, val, currency='EUR', cents=True,
```
pedrobaeza, line 20: Wasn't the problem to use this to have the word `centavo` instead of `céntimo`? This should be covered by a test.

carlosdauden: Original base method uses CURRENCY_FORMS to set this words:
https://github.com/OCA/account-payment/pull/387/files#diff-eff0847a9b2281d7070165ce134fa319d9d11497af66581c82931c8deb70446eR13-R15

pedrobaeza: OK, can you please add a test to ensure results?

carlosdauden: I was working on it. Test added.

sergio-teruel (approved):

pedrobaeza (approved):

pedrobaeza:
/ocabot merge patch

This should be fw-port to v13

PR #388, opened 2021-01-31
[13.0][IMP] account_due_list: Better UX in view, filters and action context

carlosdauden:
- Refactor tree view to add fields, use optional show and improve decorators.
 - Refactor and improve filters.
 - Add default search with new filter to avoid that unposted account move lines are showed.

@Tecnativa TT28007

### DIFF
--- a/account_due_list/views/payment_view.xml
+++ b/account_due_list/views/payment_view.xml
@@ -9,25 +9,62 @@
         <field name="arch" type="xml">
             <tree
                 string="Payments"
-                decoration-muted="reconciled!=False"
-                decoration-danger="reconciled==False and date_maturity&lt;current_date"
+                decoration-success="full_reconcile_id!=False"
+                decoration-muted="reconciled!=False and full_reconcile_id==False"
+                decoration-warning="parent_state!='posted'"
+                decoration-danger="reconciled==False and date_maturity&lt;current_date and parent_state=='posted'"
                 create="false"
                 delete="false"
             >
-                <field name="move_id" string="Invoice" readonly="1" />
-                <field name="invoice_date" readonly="1" />
-                <field name="invoice_origin" readonly="1" />
-                <field name="name" readonly="1" />
-                <field name="partner_id" readonly="1" />
-                <field name="partner_ref" readonly="1" />
-                <field name="payment_term_id" readonly="1" />
+                <field name="date" readonly="1" optional="hide" />
+                <field name="invoice_date" readonly="1" optional="show" />
+                <field
+                    name="company_id"
+                    readonly="1"
+                    groups="base.group_multi_company"
+                    optional="hide"
+                />
+                <field name="journal_id" readonly="1" optional="hide" />
+                <field name="move_id" string="Invoice" readonly="1" optional="show" />
+                <field name="invoice_origin" readonly="1" optional="show" />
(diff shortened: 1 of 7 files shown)


account_due_list/views/payment_view.xml
```diff
@@ -9,25 +9,54 @@
...
+                <field name="invoice_date" readonly="1" optional="show" />
+                <field
+                    name="company_id"
+                    readonly="1"
+                    groups="base.group_multi_company"
+                    optional="hide"
+                />
+                <field name="journal_id" readonly="1" optional="hide" />
+                <field name="move_id" string="Invoice" readonly="1" optional="show" />
+                <field name="invoice_origin" readonly="1" optional="show" />
```
pedrobaeza, line 29: Why this one is shown?

account_due_list/views/payment_view.xml
```diff
@@ -9,25 +9,54 @@
...
+                <field name="name" readonly="1" optional="show" />
+                <field name="partner_id" readonly="1" optional="show" />
+                <field name="partner_ref" readonly="1" optional="show" />
```
pedrobaeza: And this one? I think it should hide by default

account_due_list/views/payment_view.xml
```diff
@@ -9,25 +9,54 @@
...
+                    optional="hide"
+                />
+                <field
+                    name="analytic_tag_ids"
+                    readonly="1"
+                    widget="many2many_tags"
+                    groups="analytic.group_analytic_tags"
+                    optional="hide"
+                />
                 <field name="debit" readonly="1" sum="Total Debit" />
```
pedrobaeza: Not sure about these 2 being always shown. This is due list, not journal item view, so for me it's more important `balance`, `amount_currency` and `amount_residual` than these 2.

pedrobaeza (commented):
Thanks for this rework. It was pretty needed.

carlosdauden:
I agree with your comments, but I wanted to keep the original behavior so as not to disturb anyone.
Changed all except the invoice_origin field which can generate more controversy with sales orders payments flow.

pedrobaeza (approved):

pedrobaeza:
I think this should filter unposted entries directly, as they are not effects.

pedrobaeza:
@JordiBForgeFlow what do you think about this improvement?

carlosdauden:
The problem is that a journal item can be canceled or in draft and it can be reconciled.
If a journal item is canceled with the intention of modifying something and validating it again and at that moment someone is consulting effects, they may lose complete vision if they do not have the possibility to see these unposted entries.

pedrobaeza:
Then what about presenting only them apart from posted ones?

carlosdauden:
I don't understand:
https://github.com/OCA/account-payment/pull/388/files#diff-51659c468c848f543e7a5321069e828eeaaa88d80b5fb9b81ac22d066c1961feR77-R87

carlosdauden:
And only posted are showed by default:
https://github.com/OCA/account-payment/pull/388/files#diff-51659c468c848f543e7a5321069e828eeaaa88d80b5fb9b81ac22d066c1961feR229

pedrobaeza:
OK, good for me

cubells (approved):
:+1: 

LGTM

Code review and test

PR #396, opened 2021-02-12
[13.0][IMP] account_payment_return: Fill credit move line with partner

hildickethan:
Fill the move line for credit with the same partner as the debit move line

Reason for the change is to be able to concile correctly with the Spanish bank rule 43.

I don't believe this change should affect anything else that might want that field to be treated differently, however since it is to my knowledge a Spanish thing (I don't know if it's relevant in other countries), it might be better to move this change to the `l10n-spain` repo

### DIFF
--- a/account_payment_return/models/payment_return.py
+++ b/account_payment_return/models/payment_return.py
@@ -225,6 +225,7 @@ class PaymentReturn(models.Model):
         move = self.env["account.move"].create(self._prepare_return_move_vals())
         total_amount = 0.0
         all_move_lines = self.env["account.move.line"]
+        debit_move_partner = self.env["res.partner"]
         for return_line in self.line_ids:
             move_line2_vals = return_line._prepare_return_move_line_vals(move)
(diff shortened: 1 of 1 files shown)


fuentes73 (approved):

account_payment_return/models/payment_return.py
```diff
@@ -251,7 +252,10 @@ def action_confirm(self):
                 )
             extra_lines_vals = return_line._prepare_extra_move_lines(move)
             move_line_model.create(extra_lines_vals)
+            debit_move_partner = move_line2.partner_id
```
carlosdauden, line 255: What's happen if we have several return lines with different partner?

hildickethan: Right now it uses the partner of the last return line, which in hindsight is questionable, but since this is 4 months old and as far as I know there haven't been any issues with it, is having different partners a real use case? I'm not really educated on these kind of accounting things :sweat_smile:

PR #397, opened 2021-03-10
[10.0][FIX]account_check_report

AaronHForgeFlow:
If several payments are involved in the same invoice the report should show only the amount in the payment we are printing.

For example, one vendor bills for 10$ each and two payments for 5$ and 7$. The first payment is fully reconciled, the second payment is partially reconciled, 5$ to the bill and 2$ as residual amount. If we print the check report for that payment:

Previous this change:
image

After this change:
image

As you can see the payment column is confusing because it shows the total paid, including other payments, that is a little bit confusing and IMHO this approach makes more sense.

Besides that, in case when the invoice has complex payment terms, there are several payables and receivable lines by invoice. The last commit prevents to show several lines by invoices in the report. Currently it may appear several lines with the same invoice.

### DIFF
--- a/account_check_report/__init__.py
+++ b/account_check_report/__init__.py
@@ -1,1 +1,2 @@
 from . import report
+from . import models
--- a/account_check_report/__manifest__.py
+++ b/account_check_report/__manifest__.py
@@ -11,7 +11,7 @@
               "Odoo Community Association (OCA)",
     'category': 'Generic Modules/Accounting',
     'website': "https://github.com/OCA/account-payment",
-    'depends': [],
+    'depends': ["account"],
     'data': [
         'report/account_check_report.xml',
     ],
--- a/account_check_report/models/__init__.py
+++ b/account_check_report/models/__init__.py
@@ -1,0 +1,1 @@
+from . import account_payment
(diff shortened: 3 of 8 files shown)


AaronHForgeFlow:
The last commit takes into account the date of the payment in the calculation of the amount due. In my previous example the first payment comes one day before the second one, so the check report should show the amount due just after the payment is applied not on today's date:
image

The printout of the second payment will show balance due = 0 as long it is the payment that fully pays the invoice:
image

AaronHForgeFlow:
The last introduce changes to include the refunds, where those refunds were applied directly to the payment. For example:

I have a check for 7$:
   I apply 5$ to an existing invoice (invoice 1). The invoice remained open.
   Then a apply 2 $ to another invoice  (invoice 2) The invoice remained open.
   Then I issue a refund for 5$ for invoice 2, so invoice 2 is fully paid.

As a result I get:
   Invoice 1 payment is 5$, the rest is due at the payment date
   Invoice 2 payment is 2$, the rest is due at the payment date
   The refund is for -5$, nothing is due

image

AaronHForgeFlow:
cc @ForgeFlow

AaronHForgeFlow:
In case when the invoice has complex payment terms, there are several payables and receivable lines by invoice. The last commit prevents to show several lines by invoices in the report

MiquelRForgeFlow (approved):
lgtm

AaronHForgeFlow:
Do not delete branch please

AaronHForgeFlow:
This is still valid but no one cares about v10 now. I am forward porting it.

PR #399, opened 2021-03-15
[13.0][FIX] account_payment_return_import + partner_aging: Set correct journal in tests to prevent error

victoralmau:
Set correct journal in tests to prevent error.

@Tecnativa TT28755


account_payment_return_import/tests/test_import_base.py
```diff
@@ -1,4 +1,5 @@
-# Copyright 2017 David Vidal <david.vidal@tecnativa.com>
+# Copyright 2017 Tecnativa - David Vidal
+# Copyright 2021 Tecnativa - David Vidal
```
pedrobaeza: Wrong name :wink:

joao-p-marques (approved):

pedrobaeza (approved):

PR #402, opened 2021-03-22
[ADD] Optional journal for the generated account move

### LINKED ISSUE #220: Account payment return: using with non bank journal
Hi,

We are considering to use the payment  return module (very good module, btw, thanks!) with payment orders and payment modes that use the transfer account mechanism. With that mode, the payment journal is not the bank journal but typically a Misc operations journal.

Would you see a problem to allow selecting journals of type `general` on payment returns, in addition to bank journals?

enricostano:
Fix https://github.com/OCA/account-payment/issues/220


enricostano:
This change is actually not needed, I will explain in the issue.

Issue #414, opened 2021-05-20
[14.0] account_payment_term_discount: can't make batch payments


dreispt:
Odoo allows selecting multiple invoices from the same Partner, and then create a Payment for the total amount, using the Action context menu option.

When having account_payment_term_discount, this is no longer allowed.
The culprit it: https://github.com/OCA/account-payment/blob/7335b5411dc21382e3c4b84918d99dad77f957e2/account_payment_term_discount/models/account_payment.py#L93

We want to be able to have the "grouped payment" capability back.
This might mean some refactoring. 

This issue was fount when doing Vendor Bill payments, but it should also be happening for Customer Invoices too.

dreispt:
@patrickrwilson

PR #398, opened 2021-03-11
[14.0][ADD] account_payment_batch_process_discount

max3903:

PR #415, opened 2021-05-21
[DRAFT][FIX][14.0] account_payment_term_discount (Issue 414)

patrickrwilson:
Fix for Issue #414

### DIFF
--- a/account_payment_term_discount/models/account_payment.py
+++ b/account_payment_term_discount/models/account_payment.py
@@ -27,7 +27,7 @@ class AccountPaymentRegister(models.TransientModel):
             res["discount_amt"] = record.discount_amt
         return res
 
-    @api.onchange("amount", "payment_difference", "payment_date")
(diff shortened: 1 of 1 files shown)


patrickrwilson:
This PR is a work in progress as there is testing and possible changes still.

@dreispt Here is the start of the fix that I could use your assistance testing. It now should work for both a single invoice payment as well as grouped batch payments. It does NOT apply the discounts if the user does a batch payment without grouping and will process normally in that case.

Is this something we should look at adding or leave it to just work in single or grouped payments?

patrickrwilson:
This will need more design work before completing but will keep here for reference until then.

dreispt:
This is probably related to #398?

patrickrwilson:
> This is probably related to #398?

Looks like it, so perhaps this pr isn't needed anymore.

patrickrwilson:
Closing this as there is currently work being done on batch processing and another module connecting them. See issue #398

PR #391, opened 2021-02-05
[14.0][ADD] account_payment_batch_process

Hardik-OSI:
Todo

- [x] Invoice Reconciliation
- [x] pre-commit

PR #417, opened 2021-05-26
[14.0][IMP] account_payment_term_discount (Allow bulk payment if no discount)

patrickrwilson:
During bulk or grouped payments, this checks if there are any invoices with a possible discount and if exist, it stops the payment process. Otherwise the process will continue allowing for bulk or grouped payments. 

This is a simple solution that addresses issue #414

### DIFF
--- a/account_payment_term_discount/models/account_payment.py
+++ b/account_payment_term_discount/models/account_payment.py
@@ -2,8 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 from dateutil.relativedelta import relativedelta
 
-from odoo import _, api, fields, models
-from odoo.exceptions import UserError
+from odoo import api, fields, models
 
 
 class AccountPaymentRegister(models.TransientModel):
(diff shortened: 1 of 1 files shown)


patrickrwilson:
@dreispt Here is the simple solution to issue #414 for your review.

dreispt:
Hello @patrickrwilson , I just learned there is WIP to implement this support at #398, that has a dependency on #391, also undergoing fixes.
@murtuzasaleh is working on it.

patrickrwilson:
> Hello @patrickrwilson , I just learned there is WIP to implement this support at #398, that has a dependency on #391, also undergoing fixes.
> @murtuzasaleh is working on it.

@dreispt Yes, they are related however I think this simple solution still would be useful in the event a person only uses this module and not the others that are currently being worked. That way they would still be able to use the out of box batch payments. Correct me if there are other thoughts.

account_payment_term_discount/models/account_payment.py
```diff
@@ -90,14 +90,17 @@ def onchange_payment_amount(self):
...
         if any(len(active_id) != 1 for record in self):
-            # For multiple invoices, there is account.register.payments wizard
-            raise UserError(
-                _(
-                    "This method should only be called to process a "
-                    "single invoice's payment."
-                )
-            )
+            for invoice in self.line_ids:
+                if invoice.move_id.discount_amt > 0:
```
dreispt:
```suggestion
            moves = self.mapped("lines_ids.move_id").filtered("discount_amt")
            if moves:
```

account_payment_term_discount/models/account_payment.py
```diff
@@ -90,14 +90,17 @@ def onchange_payment_amount(self):
     def action_create_payments(self):
         active_id = self.env.context.get("active_ids", [])
 
+        # Don't allow bulk or grouped payments if an invoice has a possible discount.
         if any(len(active_id) != 1 for record in self):
```
dreispt:
```suggestion
        if len(active_id) > 1:
```

patrickrwilson:
@dreispt changes made, travis is red due to account_payment_return module, no errors on this module that I can see.

RLeeGML:
@patrickrwilson  The issues with account_payment_return should be resolved, if you rebase it may fix travis check

patrickrwilson:
@RLeeOSI Something is strange with my branch and it won't let me resolve conflicts because I don't have write access to my own branch, I'm going to close this and create a new one under a new branch. Sorry for the noice.

patrickrwilson:
@RLeeOSI and after looking deeper, this issue looks to have been resolved with https://github.com/OCA/account-payment/pull/398 anyway so this PR is not needed anymore. I won't be opening a new one.

PR #418, opened 2021-06-02
[14.0][ADD] account_payment_terminal: new module Account Payment Terminal

sbejaoui:
allows to pay invoices using payment terminal

https://user-images.githubusercontent.com/12665409/120458736-cf47cf80-c397-11eb-8c79-28719d65d374.mp4

### DIFF
--- a/account_payment_terminal/models/account_payment_terminal.py
+++ b/account_payment_terminal/models/account_payment_terminal.py
@@ -8,7 +8,6 @@ class AccountPaymentTerminal(models.Model):
 
     _name = "account.payment.terminal"
     _description = "Account Payment Terminal"
-    _inherit = "mail.thread"
 
     name = fields.Char(required=True)
     proxy_ip = fields.Char(
--- a/account_payment_terminal/models/account_payment_terminal.py
+++ b/account_payment_terminal/models/account_payment_terminal.py
@@ -1,0 +1,26 @@
+# Copyright 2021 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class AccountPaymentTerminal(models.Model):
+
+    _name = "account.payment.terminal"
+    _description = "Account Payment Terminal"
+    _inherit = "mail.thread"
+
+    name = fields.Char(required=True)
+    proxy_ip = fields.Char(
+        string="IP Address",
+        size=45,
+        help="The hostname or ip address of the hardware proxy",
+        required=True,
+    )
+    oca_payment_terminal_id = fields.Char(
+        string="Terminal identifier",
+        help=(
+            "The identifier of the terminal as known by the hardware proxy. "
+            "Leave empty if the proxy has only one terminal connected."
+        ),
+    )
--- a/account_payment_terminal/models/oca_payment_terminal_form_mixin.py
+++ b/account_payment_terminal/models/oca_payment_terminal_form_mixin.py
@@ -62,11 +62,7 @@ class OcaPaymentTerminalFormMixin(models.AbstractModel):
 
     # To define in the heir model
     def get_payment_info(self):
-        return {
-            "proxy_ip": self.account_payment_terminal_id.proxy_ip,
-            "oca_payment_terminal_id": self.account_payment_terminal_id.oca_payment_terminal_id,
-            "payment_mode": "card",  # TODO: Add check mode?
-        }
+        raise NotImplementedError()
 
     def action_confirm_payment(self, payment_reference):
         raise NotImplementedError()
--- a/account_payment_terminal/models/oca_payment_terminal_form_mixin.py
+++ b/account_payment_terminal/models/oca_payment_terminal_form_mixin.py
@@ -1,0 +1,72 @@
+# Copyright 2021 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class OcaPaymentTerminalFormMixin(models.AbstractModel):
+
+    _name = "oca.payment.terminal.form.mixin"
+    _description = "OCA Payment Terminal Form Mixin"
+
+    account_payment_terminal_id = fields.Many2one(
+        comodel_name="account.payment.terminal", string="Payment Terminal"
+    )
+
+    def action_payment_terminal_transaction_start(self):
+        self.ensure_one()
+        if not hasattr(self, "journal_id"):
+            # the object must be linked to journal in order to get the payment
+            # terminal type
+            return {}
+        if (
+            self.journal_id.use_payment_terminal
+            and self.account_payment_terminal_id
+            and hasattr(
+                self, self._get_payment_terminal_transaction_start_method_name()
+            )
+        ):
+            return getattr(
+                self, self._get_payment_terminal_transaction_start_method_name()
+            )()
+        raise UserError(
+            _(
+                "Payment terminal is not correctly configured. "
+                "Please contact your administrator."
+            )
+        )
+
+    def _get_payment_terminal_transaction_start_method_name(self):
+        return "_%s_transaction_start" % self.journal_id.use_payment_terminal or ""
+
+    def _oca_payment_terminal_transaction_start(self):
+        self.ensure_one()
+        view_id = self._get_payment_terminal_form_view_id()
+        return {
+            "name": _("Payment Terminal"),
+            "type": "ir.actions.act_window",
+            "view_mode": "oca_payment_terminal_form",
+            "res_model": self._name,
+            "res_id": self.id,
+            "target": "new",
+            "view_id": view_id,
+            "context": self.env.context,
+        }
+
+    @api.model
+    def _get_payment_terminal_form_view_id(self):
+        return self.env.ref(
+            "account_payment_terminal.oca_payment_terminal_form_mixin_form_view"
+        ).id
+
+    # To define in the heir model
+    def get_payment_info(self):
+        return {
+            "proxy_ip": self.account_payment_terminal_id.proxy_ip,
+            "oca_payment_terminal_id": self.account_payment_terminal_id.oca_payment_terminal_id,
+            "payment_mode": "card",  # TODO: Add check mode?
+        }
+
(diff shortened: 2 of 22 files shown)


sbejaoui:
cc/ @sbidoul

rousseldenis (approved):
Code review. LGTM

@sbejaoui Maybe you can add the payment video in PR ?

account_payment_terminal/models/account_payment_terminal.py
```diff
@@ -0,0 +1,26 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class AccountPaymentTerminal(models.Model):
+
+    _name = "account.payment.terminal"
+    _description = "Account Payment Terminal"
+    _inherit = "mail.thread"
```
sbidoul: Is mail.thread useful for configuration data ?

account_payment_terminal/models/oca_payment_terminal_form_mixin.py
```diff
@@ -0,0 +1,72 @@
...
+            )()
+        raise UserError(
+            _(
+                "Payment terminal is not correctly configured. "
+                "Please contact your administrator."
+            )
+        )
+
+    def _get_payment_terminal_transaction_start_method_name(self):
+        return "_%s_transaction_start" % self.journal_id.use_payment_terminal or ""
```
sbidoul, line 41: Is that not too much configurability ?

account_payment_terminal/models/oca_payment_terminal_form_mixin.py
```diff
@@ -0,0 +1,72 @@
...
+    # To define in the heir model
+    def get_payment_info(self):
+        return {
+            "proxy_ip": self.account_payment_terminal_id.proxy_ip,
+            "oca_payment_terminal_id": self.account_payment_terminal_id.oca_payment_terminal_id,
+            "payment_mode": "card",  # TODO: Add check mode?
+        }
+
+    def action_confirm_payment(self, payment_reference):
+        pass
```
sbidoul:
```suggestion
        raise NotImplementedError()
```

account_payment_terminal/models/oca_payment_terminal_form_mixin.py
```diff
@@ -0,0 +1,72 @@
...
+            "account_payment_terminal.oca_payment_terminal_form_mixin_form_view"
+        ).id
+
+    # To define in the heir model
+    def get_payment_info(self):
+        return {
+            "proxy_ip": self.account_payment_terminal_id.proxy_ip,
+            "oca_payment_terminal_id": self.account_payment_terminal_id.oca_payment_terminal_id,
+            "payment_mode": "card",  # TODO: Add check mode?
+        }
```
sbidoul: Here I'd call another method that provides the real payment information and `raise NotImplementedError()` in the mixin. It will be more obvious that it needs to be overridden.

account_payment_terminal/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This addon allows to pay invoices using payment terminal.
```
sbidoul:
```suggestion
This addon allows to pay invoices using a payment terminal. It communicates with the payment terminal proxy using the same protocol as OCA `pos_payment_terminal`.
```

account_payment_terminal/readme/USAGE.rst
```diff
@@ -0,0 +1,5 @@
+#. Go to Invoicing > Customers > Invoices and select an open invoice.
+#. Register a payment.
+#. Select the bank journal configured to use payment terminals.
+#. Select a terminal.
+#. Confirm.
```
sbidoul, line 5: Maybe mention that this can be tested using the mock terminal of pywebdriver (plus link)?

account_payment_terminal/wizards/account_payment_register.py
```diff
@@ -0,0 +1,41 @@
...
+
+    need_payment_terminal = fields.Boolean(compute="_compute_need_payment_terminal")
+
+    @api.depends("journal_id")
+    def _compute_need_payment_terminal(self):
+        for rec in self:
+            if rec.journal_id.use_payment_terminal:
+                rec.need_payment_terminal = True
+            else:
+                rec.need_payment_terminal = False
```
sbidoul:
```suggestion
            rec.need_payment_terminal = rec.journal_id.use_payment_terminal
```

?

account_payment_terminal/wizards/account_payment_register.py
```diff
@@ -0,0 +1,41 @@
...
+        self.update({"account_payment_terminal_id": False})
+
+    def get_payment_info(self):
+        res = super().get_payment_info()
+        res.update(
+            {
+                "amount": self.amount,
+                "currency_iso": self.currency_id.name,
+                "currency_decimals": self.currency_id.decimal_places,
+                "payment_id": self.id,
```
sbidoul: where is this `payment_id` field used ?

sbidoul (commented):
Nice!

sbidoul:
Too bad the video does not render ? Could you try another format ?

sbidoul:
@legalsylvain @PierrickBrun you may be interested in this one: pay invoices from the Odoo backend using a payment terminal, with the same protocol as pos_payment_terminal (including capture of the transaction reference for easy reconciliation).

account_payment_terminal/models/account_payment_terminal.py
sbejaoui: I was thinking in case user needs to keep notes for himself or other technical users. But not really important.

account_payment_terminal/models/oca_payment_terminal_form_mixin.py
sbejaoui: make sense

account_payment_terminal/wizards/account_payment_register.py
sbejaoui: `use_payment_terminal` is a selection field, the ORM will convert `need_payment_terminal` to boolean after assigning a string?

legalsylvain:
I'm not sure understanding the use case. You have some customers with payment terminals, without the point of sale module installed ?

account_payment_terminal/wizards/account_payment_register.py
sbidoul: Ah, yes, it should convert (or you can write `bool(rec.journal_id.use_payment_terminal)` to be explicit). But from the variable name I honestly thought `use_payment_terminal` was a bool :)

sbidoul:
> I'm not sure understanding the use case. You have some customers with payment terminals, without the point of sale module installed ?

Yes: sales.wo.men at their desk selling furniture, and asking customer for partial payment. Integrating with the terminal and capturing the payment reference allows for easier payment reconciliation.

account_payment_terminal/models/oca_payment_terminal_form_mixin.py
sbejaoui: If we wont to keep one action for payment start (for simpler view) we should get the specific one dynamically.

legalsylvain:
> Yes: sales.wo.men at their desk selling furniture, and asking customer for partial payment. Integrating with the terminal and capturing the payment reference allows for easier payment reconciliation.

Well, I see. 
I don't use 14.0 for the time being, so maybe it's not relevant, but in V12, we have two modules with similar features ``account_cash_invoice`` (not very good named BTW) [here](https://github.com/OCA/account-payment/tree/12.0/account_cash_invoice) and ``pos_session_pay_invoice`` [here](https://github.com/OCA/pos/tree/12.0/pos_session_pay_invoice) that allows to reconcile invoices with pos payment.

if we make a front-end interface for the ``pos_session_pay_invoice`` I think it covers all the features of the current module.
- connexion with payment terminal
- mark invoiced as paid
- reconcile

Don't you think ?

account_payment_terminal/wizards/account_payment_register.py
sbejaoui: https://github.com/OCA/pos/blob/14.0/pos_payment_terminal/static/src/js/payment_terminal.js#L43

sbidoul:
Yeah we also made a `pos_pay_invoice` for 10.0 [here](https://github.com/acsone/pos/tree/10.0-pos_pay_invoice-bwi/pos_pay_invoice) but it was complex and we hope to get rid of it with this. We'll look at yours though.

PierrickBrun:
I like the way it is done here. I think it is better to be able to edit the invoice in the backend and then pay it right after.

I think it could also be used in a context where someone buys something online then adds something to the order while in the store.

account_payment_terminal/static/src/js/views/payment_terminal_form/payment_terminal_form_controller.js
```diff
@@ -0,0 +1,160 @@
...
+    var OCAPaymentTerminalFormController = FormController.extend({
+        init: function () {
+            this.transaction_id = null;
+            this.oca_payment_terminal_id = null;
+            this.proxy_url = null;
+            this._super.apply(this, arguments);
+        },
+        _show_error: function (error_message) {
+            this.$el.find("p").text(error_message);
+            this.$el.find("p").css("color", "red");
```
MeetKD: Suggestion : remove this line and add css attribute in the first place(error msg)

account_payment_terminal/static/src/js/views/payment_terminal_form/payment_terminal_form_controller.js
```diff
@@ -0,0 +1,160 @@
...
+            console.error(error_message);
+        },
+        _show_success: function (success_message) {
+            this.$el.find("p").text(success_message);
+            this.$el.find("p").css("color", "green");
```
MeetKD: Same here

sbejaoui:
Hi @dreispt ,

Why the label `need fixing`?

dreispt:
I believe there are comments requesting changes. Let me know if this is incorrect.

rousseldenis:
> I believe there are comments requesting changes. Let me know if this is incorrect.

They are suggestions from @MeetKD, not requirements I think

adrienpeiffer:
@dreispt We just fixed last comments from @MeetKD

adrienpeiffer (approved):

Issue #283, opened 2019-09-09
[account_payment_return] Cannot reconcile with bank statement


jalzaga:
When we create a payment return related to an invoice paid through a debit order, it creates the corresponding journal entry and it lets the invoice unpaid. The problem is that you cannot reconcile de unpaid amount with the bank statement because it is not shown in blue as the original payment.
To recreate the situation just create an invoice with a direct debit payment mode an include it in a debit order. The payment mode is set with offsetting_account = bank account
You can confirm the debit order so that the invoice becomes paid and you can reconcile the payment with the bank statement because it is shown in blue at the reconciliation screen.
Then you create a payment return for that invoice and confirm it, so that a journal entry is created. The account move contains the credit to the bank account but it does not appear in the reconciliation screen. I guess because there is no bank payment line create for that return.
I tested in runbot version 11.

jalzaga:
There is no answer to this issue since September, 9. There are customers waiting for a solution to it. We know there is a workaround using a transfer account but it is not the best.
Please give feedback about it.

rafaelbn:
Hi @jalzaga ,

In which version?

I don't have any issue.

Maybe you can make a PR to solve it?

jalzaga:
I tested in v11 but I guess its also in v10 and v12 @rafaelbn

jalzaga:
@rafaelbn is there any news about this issue?

enricostano:
@jalzaga do you think that this issue is related to #220 ?

PR #420, opened 2021-06-25
[IMP] account_payment_return: Add general type journals to be selected in payment return

carlosdauden:
This patch allows the use of general type journals for the creation of the payment return voucher.

- https://github.com/OCA/account-payment/issues/220
- https://github.com/OCA/account-payment/issues/283

Regardless of this, we should make the appropriate modifications in the modules dependent of `account_payment_order` so that they adapt to the new flow and use the fields `payment_debit_account_id` `payment_credit_account_id` and `suspense_account_id`.

@Tecnativa TT29784

### DIFF
--- a/account_payment_return/views/payment_return_view.xml
+++ b/account_payment_return/views/payment_return_view.xml
@@ -51,7 +51,7 @@
                         <field name="name" />
                         <field
                             name="journal_id"
-                            domain="[('type', 'in', ['bank', 'cash'])]"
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

pedrobaeza:
@enricostano @sbidoul please confirm according conversation in #220

sbidoul (approved):

pedrobaeza:
Thanks

/ocabot merge patch

PR #421, opened 2021-06-25
[12.0][ADD] account_payment_term_partner_holiday: Add partner holidays to exclude it when payment term assigned in invoices set due dates

victoralmau:
This addon adds the possibility of defining holiday periods in a partner so as not to use those periods as the due date on invoices.

@Tecnativa TT30526


account_payment_term_partner_holiday/README.rst
```diff
@@ -0,0 +1,95 @@
...
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to 'Contacts' and create or edit some record.
+#. Go to 'Sales & Purchases' tab and create some holidays records.
+#. Go to 'Invoicing > Customers > Invoices' or to 'Invoicing > Vendors >
+   Invoices' and create or edit some record.
+#. If apply some payment terms and Due date is changed not use holidays date.
```
pedrobaeza:
```suggestion
#. If the computed due date is inside a holidays period, it's moved to the first available date.
```

account_payment_term_partner_holiday/README.rst
```diff
@@ -0,0 +1,95 @@
...
+~~~~~~~
+
+* Tecnativa
+
+Contributors
+~~~~~~~~~~~~
+
+* `Tecnativa <https://www.tecnativa.com>`__:
+
+  * Víctor Martínez
```
pedrobaeza, line 70: Should you include me as designer/reviewer?

account_payment_term_partner_holiday/models/res_partner_holiday.py
```diff
@@ -0,0 +1,88 @@
...
+            ("23", "23"),
+            ("24", "24"),
+            ("25", "25"),
+            ("26", "26"),
+            ("27", "27"),
+            ("28", "28"),
+            ("29", "29"),
+            ("39", "30"),
+            ("31", "31")
+        ]
```
pedrobaeza:
```suggestion
        return [(str(i), str(i)) for i in range(1, 32)]
```

account_payment_term_partner_holiday/models/res_partner_holiday.py
```diff
@@ -0,0 +1,88 @@
...
+
+    partner_id = fields.Many2one(
+        comodel_name="res.partner",
+        string="Partner",
+        required=True,
+        ondelete="cascade",
+        index=True,
+        copy=False
+    )
+    sequence = fields.Integer(string="Sequence", default=10)
```
pedrobaeza: It shouldn't have sequence, but ordered by starting month, day

account_payment_term_partner_holiday/models/res_partner_holiday.py
```diff
@@ -0,0 +1,88 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import _, api, fields, models
+
+
+class ResPartnerHoliday(models.Model):
+    _name = "res.partner.holiday"
+    _description = "Partner Holidays"
```
pedrobaeza:
```suggestion
    _description = "Partner Holidays"
    _order = "month_from, day_from, id"
```

account_payment_term_partner_holiday/models/res_partner_holiday.py
```diff
@@ -0,0 +1,88 @@
...
+
+    @api.model
+    def _get_days(self):
```
pedrobaeza: Follow guidelines for method name (for both):
```suggestion
    def _selection_days(self):
```

account_payment_term_partner_holiday/models/res_partner_holiday.py
```diff
@@ -0,0 +1,88 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
```
pedrobaeza: As this model is a one2many of res.partner, put it in the same file.

account_payment_term_partner_holiday/models/res_partner.py
```diff
@@ -0,0 +1,24 @@
...
+    holiday_ids = fields.One2many(
+        comodel_name="res.partner.holiday",
+        inverse_name="partner_id",
+        string="Holidays",
+        copy=True,
+        auto_join=True
+    )
+
+    def is_date_in_holiday_ids(self, date):
+        if isinstance(date, str):
```
pedrobaeza, line 20: This is not happening in v12 AFAIK, so no need to add this condition. Anyways, it should be `fields.Date`, not `fields.Datetime`.

account_payment_term_partner_holiday/security/security.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="rule_res_partner_holiday" model="ir.rule">
+        <field name="name">Res partner holiday</field>
+        <field name="model_id" ref="model_res_partner_holiday" />
+        <field name="global" eval="True" />
+        <field name="domain_force">[(1, '=', 1)]</field>
```
pedrobaeza: Adding this serves for nothing. No record rules = full access.

account_payment_term_partner_holiday/views/res_partner_view.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
+<odoo>
+    <record id="view_res_partner_form_holiday_ids" model="ir.ui.view">
+        <field name="name">res.partner.holiday_ids</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form"/>
+        <field name="arch" type="xml">
+            <field name="industry_id" position="after">
+                <field name="holiday_ids">
```
pedrobaeza, line 9: Put this under `payment_term_id`.

account_payment_term_partner_holiday/models/account_invoice.py
```diff
@@ -0,0 +1,28 @@
...
+
+from odoo import api, models
+
+
+class AccountInvoice(models.Model):
+    _inherit = "account.invoice"
+
+    @api.onchange("date_due")
+    def _onchange_date_due(self):
+        if self.date_due and self.payment_term_id and self.partner_id:
```
pedrobaeza, line 20: This is not the goal of the module (although it can be covered as well), but to modify the date when using payment terms, that make use of this method:

https://github.com/odoo/odoo/blob/3353fef07a3a54596b30aabc29d9729f2eb3875e/addons/account/models/account_invoice.py#L2057

The best technique is to call super, and the goes through each returned line for modifying forbidden dates.

account_payment_term_partner_holiday/models/account_invoice.py
```diff
@@ -0,0 +1,28 @@
...
+            new_date_due = self.date_due
+            is_date_in_holiday = self.partner_id.is_date_in_holiday_ids(new_date_due)
+            pterm = self.payment_term_id.with_context(
+                currency_id=self.company_id.currency_id.id
+            )
+            while is_date_in_holiday:
```
pedrobaeza: Instead of this loop, I think the called method should return the forbidden interval (or False if not), and then put following day of the end date.

account_payment_term_partner_holiday/models/res_partner.py
```diff
@@ -0,0 +1,24 @@
...
+
+    holiday_ids = fields.One2many(
+        comodel_name="res.partner.holiday",
+        inverse_name="partner_id",
+        string="Holidays",
+        copy=True,
+        auto_join=True
+    )
+
+    def is_date_in_holiday_ids(self, date):
```
pedrobaeza: Why putting `_ids` suffix?

pedrobaeza (changes requested):

account_payment_term_partner_holiday/tests/test_partner_holiday.py
```diff
@@ -0,0 +1,98 @@
...
+from odoo.tests import common
+from odoo.tests.common import Form
+
+
+class TestPartnerHoliday(common.TransactionCase):
+    def setUp(self):
+        super().setUp()
+        self.partner_1 = self.env["res.partner"].create({"name": "Partner test 1"})
+        self.partner_2 = self.env["res.partner"].create({"name": "Partner test 2"})
+        self.partner_1.write({
```
joao-p-marques: Why not putting this in the creation above? If there is this particular need, maybe put a comment here explaining it to make it clearer...

joao-p-marques (commented):
LGTM, apart from @pedrobaeza's suggestions

Just a minor comment

victoralmau:
Thanks @pedrobaeza and @joao-p-marques, changes done.

joao-p-marques (approved):

account_payment_term_partner_holiday/models/res_partner.py
```diff
@@ -0,0 +1,87 @@
...
+        if isinstance(date, str):
+            date = fields.Date.from_string(date)
+        res = self.holiday_ids.filtered(
+            lambda x: int(x.month_from) <= date.month and int(x.month_to) >= date.month
+            and int(x.day_from) <= date.day and int(x.day_to) >= date.day
+        )
+        if bool(res):
+            res = res[0]
+            return [
+                fields.Date.from_string(
```
pedrobaeza, line 38: One of this two can give an error when doing the conversion, so you should sanitize returning the minimum allowed date. Example: you put 31 of February, you should return 28 or 29 depending on the year. It can be an iterating method that substract one while `from_string` gives error.

account_payment_term_partner_holiday/models/account_payment_term.py
```diff
@@ -0,0 +1,29 @@
...
+from dateutil.relativedelta import relativedelta
+
+
+class AccountPaymentTerm(models.Model):
+    _inherit = "account.payment.term"
+
+    @api.one
+    def compute(self, value, date_ref=False):
+        result = super().compute(value=value, date_ref=date_ref)[0]
+        if self.env.context.get("default_partner_id"):
```
pedrobaeza: Is this context the proper one? I think you should inject an specific context from invoices when doing this computation, as I think it's not assured to have this context always (think also on other automations like sales invoicing).

account_payment_term_partner_holiday/models/account_payment_term.py
```diff
@@ -0,0 +1,29 @@
...
+                self.env.context.get("default_partner_id")
+            )
+            result2 = []
+            for item in result:
+                date_item = item[0]
+                is_date_in_holiday = partner.is_date_in_holiday(date_item)
+                if is_date_in_holiday:
+                    next_date = is_date_in_holiday[1]
+                    next_date += relativedelta(days=1)
+                    result2.append((fields.Date.to_string(next_date), item[0]))
```
pedrobaeza: Shouldn't it be?

```suggestion
                    result2.append((fields.Date.to_string(next_date), item[1]))
```

account_payment_term_partner_holiday/models/res_partner.py
victoralmau: Ok, i have change it and add February in test to test it.

account_payment_term_partner_holiday/models/account_payment_term.py
victoralmau: You are right, the name of the context is not the most appropriate but it is something that is launched automatically (https://github.com/odoo/odoo/blob/3353fef07a3a54596b30aabc29d9729f2eb3875e/addons/account/models/account_invoice.py#L2057). You mean `_onchange_payment_term_id()` should be overwritten and add custom context to use in `compute()` function?
I think it is important to note that this function and `_onchange_date_due()` do "the same".

pedrobaeza: I mean that you inject a specific context key in the methods that call this one:

https://github.com/odoo/odoo/blob/a2463b213e7f19b4a089707b226f9a52b4a52b7a/addons/account/models/account_invoice.py#L908
https://github.com/odoo/odoo/blob/a2463b213e7f19b4a089707b226f9a52b4a52b7a/addons/account/models/account_invoice.py#L1301

so you assure that you don't depend on external agents.

account_payment_term_partner_holiday/models/res_partner.py
```diff
@@ -0,0 +1,93 @@
...
+        string="Month from"
+    )
+    day_to = fields.Selection(
+        selection="_selection_days",
+        string="Day to"
+    )
+    month_to = fields.Selection(
+        selection="_selection_months",
+        string="Month to"
+    )
```
pedrobaeza, line 98: Add a constraint for not setting the end before the start

pedrobaeza (changes requested):
Travis is red

account_payment_term_partner_holiday/models/res_partner.py
```diff
@@ -0,0 +1,107 @@
...
+
+    @api.constrains("day_from", "month_from", "day_to", "month_to")
+    def _check_from_end_dates(self):
+        for item in self:
+            if (int(item.month_from) > int(item.month_to)) or (
+                int(item.month_to) == int(item.month_from)
+                and int(item.day_from) > int(item.day_to)
+            ):
+                raise ValidationError(_("Date to is need to set after date from"))
```
pedrobaeza: Please test several of these cases. A proper redaction for your sentence would be: `Date to is needed to be set after date from`, but the redaction is still confusing, so I recommend:
```suggestion
                raise ValidationError(_("You can't set the ending holidays period before the beginning."))
```

account_payment_term_partner_holiday/models/account_payment_term.py
pedrobaeza: This is still pending.

account_payment_term_partner_holiday/models/res_partner.py
```diff
@@ -0,0 +1,107 @@
...
+from odoo.exceptions import ValidationError
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    holiday_ids = fields.One2many(
+        comodel_name="res.partner.holiday",
+        inverse_name="partner_id",
+        string="Holidays",
```
pedrobaeza: We need to be more explicit:

```suggestion
        string="Payment Terms Holidays",
```

pedrobaeza (changes requested):

victoralmau:
Thanks @pedrobaeza. changes done.

account_payment_term_partner_holiday/models/account_invoice.py
```diff
@@ -0,0 +1,33 @@
...
+
+from odoo import api, models
+from dateutil.relativedelta import relativedelta
+
+
+class AccountInvoice(models.Model):
+    _inherit = "account.invoice"
+
+    @api.onchange('payment_term_id', 'date_invoice')
+    def _onchange_payment_term_date_invoice_account_payment_term_partner_holiday(self):
```
pedrobaeza: In this you mustn't change onchange name. You only need to do it when the module that adds the field doesn't declare an onchange for that field.

account_payment_term_partner_holiday/models/account_invoice.py
```diff
@@ -0,0 +1,33 @@
...
+            new_date_due = self.date_due
+            is_date_in_holiday = self.partner_id.is_date_in_holiday(new_date_due)
+            if is_date_in_holiday:
+                res_date_end = is_date_in_holiday[1]
+                new_date_due = res_date_end + relativedelta(days=1)
+            if new_date_due != self.date_due:
+                self.date_due = new_date_due
+
+    def action_move_create(self):
+        """ Is not possible call super in multi because we need to add partner_id
```
pedrobaeza: Explain what and why you do this:

```suggestion
        """Inject a context for getting the partner when computing payment term.
        The trade-off is that we should split the call to super record per record,
        but it shouldn't impact in performance.
        """
```

pedrobaeza (approved):

PR #425, opened 2021-07-05
[13.0][MIG] account_payment_term_partner_holiday: Migration to 13.0

victoralmau:
Migration to 13.0.

@Tecnativa TT30526


pedrobaeza (approved):

dreispt (approved):

account_payment_term_partner_holiday/models/res_partner.py
```diff
@@ -34,7 +34,7 @@ def is_date_in_holiday(self, date):
                     res_date_to = fields.Date.from_string(
                         "%s-%s-%s" % (date.year, res.month_to, day_to)
                     )
-                except:
+                except BaseException:
```
joao-p-marques: As we are changing this, could this maybe be a more specific exception? I guess this should probably come from the conversion from `string` to `Date`, so you probably want to catch a `ValueError`: https://docs.python.org/3/library/datetime.html#datetime.datetime.strptime

joao-p-marques (changes requested):
Code and functional review :+1: 

The only thing that is not so clear is that, at least in v13, the computed due date is not shown, so the user has no idea of what the result is (I don't remember if this was also the case for v12). It happens because Odoo hides that field if a payment term is set, so maybe it would make sense to force its visibility here? Something similar is done in https://github.com/OCA/account-invoicing/tree/13.0/account_invoice_date_due, that allows that field to be editable

pedrobaeza:
No need to show the computed due date. That's out of the scope of the module.

PR #426, opened 2021-07-05
[12.0][FIX] account_payment_term_partner_holiday: Prevent error in tests when sale addon installed

victoralmau:
Prevent error in tests when sale addon installed.

Steps to reproduce:
- Install `account_payment_term_partner_holiday`
- Install `sale`
- Run `account_payment_term_partner_holiday` tests

@Tecnativa TT30526

### DIFF
--- a/account_payment_term_partner_holiday/tests/test_partner_holiday.py
+++ b/account_payment_term_partner_holiday/tests/test_partner_holiday.py
@@ -6,6 +6,8 @@ from odoo.tests.common import Form
 from odoo import fields
 
 
+@common.at_install(False)
+@common.post_install(True)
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This is not the solution at all, as this is only postponing other kind of problems. Please start putting which is the error in integration environments, and then we can see options.

victoralmau:
> This is not the solution at all, as this is only postponing other kind of problems. Please start putting which is the error in integration environments, and then we can see options.

Description explain now steps to reproduce error in tests.

pedrobaeza:
I still don't see the error given in tests.

victoralmau:
Sorry, after careful review it seems that no change is necessary.

PR #427, opened 2021-07-08
[13.0][FIX] account_check_printing_report_base: Pin num2words version and fix typo in function call

joao-p-marques:
The current version used and installed by Odoo is 0.5.6: https://github.com/odoo/odoo/blob/13.0/requirements.txt#L21
It has a typo in the function call to `to_currency` where "separator" is spelled "seperator".
As that is the version that Odoo uses, we should adapt to it, and also pin to ensure we are installing the correct one.

The typo is fixed in https://github.com/savoirfairelinux/num2words/commit/2eee037648eb2245d9714f694e4868b80082a216 for version 0.5.10, so we should revert this if we ever upgrade to that version

@Tecnativa
TT30932

ping @pedrobaeza @sergio-teruel @CarlosRoca13

### DIFF
--- a/account_check_printing_report_base/report/lang.py
+++ b/account_check_printing_report_base/report/lang.py
@@ -1,4 +1,3 @@
-# pylint: disable=missing-import-error
 from num2words import CONVERTER_CLASSES, CONVERTES_TYPES, num2words
 from num2words.base import Num2Word_Base
 from num2words.lang_ES import Num2Word_ES
@@ -12,9 +11,15 @@ class Num2WordESCustom(Num2Word_ES):
 
     # TODO: PR to remove overwrite in num2words code and use CURRENCY_FORMS
     def to_currency(self, val, longval=True, old=False):
-        return Num2Word_Base.to_currency(
-            self, val, currency="EUR", cents=True, separator=" con", adjective=False
-        )
+        try:
+            # On older versions we need to use "seperator" instead of "separator"
+            return Num2Word_Base.to_currency(
+                self, val, currency="EUR", cents=True, seperator=" con", adjective=False
+            )
+        except TypeError:
+            return Num2Word_Base.to_currency(
+                self, val, currency="EUR", cents=True, separator=" con", adjective=False
+            )
 
 
 def num2words_custom(number, ordinal=False, lang="en", to="cardinal", **kwargs):
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 # generated from manifests external_dependencies
-num2words==0.5.10
+num2words


carlosdauden:
ping @i-vyshnevska @bodedra 
Is it right for you?

pedrobaeza:
Instead of changing these things, do the call 2 times with one or the other argument enclosed in a try, and let the deployment to decide which version wins.

joao-p-marques:
> Instead of changing these things, do the call 2 times with one or the other argument enclosed in a try, and let the deployment to decide which version wins.

My latest change should make it work in both versions, although IMHO it would make sense to stick with `0.5.6` where possible, to avoid compatibility issues with Odoo.

account_check_printing_report_base/__manifest__.py
```diff
@@ -16,7 +16,7 @@
     "category": "Generic Modules/Accounting",
     "website": "https://github.com/OCA/account-payment",
     "depends": ["account_check_printing", "report_wkhtmltopdf_param"],
-    "external_dependencies": {"python": ["num2words"]},
+    "external_dependencies": {"python": ["num2words==0.5.6"]},
```
pedrobaeza: Please remove this

account_check_printing_report_base/report/lang.py
```diff
@@ -12,9 +14,20 @@ class Num2WordESCustom(Num2Word_ES):
 
     # TODO: PR to remove overwrite in num2words code and use CURRENCY_FORMS
     def to_currency(self, val, longval=True, old=False):
-        return Num2Word_Base.to_currency(
-            self, val, currency="EUR", cents=True, separator=" con", adjective=False
-        )
+        if StrictVersion(
```
pedrobaeza: I prefer the `try` approach, as you are still sticking to something incorrect: what would happen on `0.5.11` version?

joao-p-marques: Well, that can also have some caveats, as I ended up doing a small refactor that I tried to explain in the comments.

Prior to version `0.5.10`, besides the typo, the problem was that the class `Num2Word_ES` (the one we inherit here) in turn inherited from the base class `Num2Word_Base` but overwrote the `to_currency` method. That is why, instead of just applying plain Python inheritance, we needed to call the base class directly with `Num2Word_Base.to_currency`.

Now the code has been standardize after https://github.com/savoirfairelinux/num2words/pull/167 and inheritance is now in place as normal, so we can call `super` directly. 

The fact that there are these 2 changes in behavior makes it necessary to work differently according to the version, not only based on the exception thrown. I think this is clearer, but I can always avoid this small refactor and use the old implementation in both situations.

> what would happen on `0.5.11` version?

Assuming the code doesn't suffer massive changes upstream, we would use the new implementation, that I mention above. I can always constraint the version to `<=0.5.10`, which is safer...

account_check_printing_report_base/__manifest__.py
joao-p-marques: Should I change it to `num2words<=0.5.10`, which the current one and the one we are effectively supporting here?

account_check_printing_report_base/report/lang.py
pedrobaeza: I really don't see the problem in the try approach and this code is far more convoluted than the other, and we shouldn't pin and library version.

pedrobaeza (approved):

joao-p-marques:
@pedrobaeza changes made :+1:

sergio-teruel (approved):

PR #429, opened 2021-07-12
[12.0][FIX] account_payment_term_partner_holiday: Improve is_date_in_holiday() to filtered correctly when day_to is less than date to compare

victoralmau:
Improve `is_date_in_holiday()` to filtered correctly when day_to is less than date to compare.
Improve is_date_in_holiday() to work fine in child partners
Hide holidays in partner view to child partners.
Change `month_from` and `month_to` fields to convert in 2 characters to order correctly holidays + Script migration

It's need to change in 13.0 too.

@Tecnativa TT30526

### DIFF
--- a/account_payment_term_partner_holiday/models/account_invoice.py
+++ b/account_payment_term_partner_holiday/models/account_invoice.py
@@ -2,7 +2,6 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 from odoo import api, models
-from dateutil.relativedelta import relativedelta
 
 
 class AccountInvoice(models.Model):
@@ -17,12 +16,8 @@ class AccountInvoice(models.Model):
 
     @api.onchange("date_due")
     def _onchange_date_due_account_payment_term_partner_holiday(self):
-        if self.date_due and self.payment_term_id and self.partner_id:
-            new_date_due = self.date_due
-            is_date_in_holiday = self.partner_id.is_date_in_holiday(new_date_due)
-            if is_date_in_holiday:
-                res_date_end = is_date_in_holiday[1]
-                new_date_due = res_date_end + relativedelta(days=1)
+        if self.date_due and self.partner_id:
+            new_date_due = self.partner_id._get_valid_due_date(self.date_due)
             if new_date_due != self.date_due:
                 self.date_due = new_date_due
 
--- a/account_payment_term_partner_holiday/models/res_partner.py
+++ b/account_payment_term_partner_holiday/models/res_partner.py
@@ -61,15 +61,15 @@ class ResPartnerHoliday(models.Model):
     @api.model
     def _selection_months(self):
         return [
-            ("1", _("January")),
-            ("2", _("February")),
-            ("3", _("March")),
-            ("4", _("April")),
-            ("5", _("May")),
-            ("6", _("June")),
-            ("7", _("July")),
-            ("8", _("August")),
-            ("9", _("September")),
+            ("01", _("January")),
+            ("02", _("February")),
+            ("03", _("March")),
+            ("04", _("April")),
+            ("05", _("May")),
+            ("06", _("June")),
+            ("07", _("July")),
+            ("08", _("August")),
+            ("09", _("September")),
             ("10", _("October")),
             ("11", _("November")),
             ("12", _("December"))
--- a/account_payment_term_partner_holiday/models/res_partner.py
+++ b/account_payment_term_partner_holiday/models/res_partner.py
@@ -1,8 +1,11 @@
 # Copyright 2021 Tecnativa - Víctor Martínez
+# Copyright 2021 Tecnativa - João Marques
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 from odoo import _, api, fields, models
+import calendar
 from odoo.exceptions import ValidationError
+from dateutil.relativedelta import relativedelta
 
 
 class ResPartner(models.Model):
@@ -16,30 +19,34 @@ class ResPartner(models.Model):
         auto_join=True
     )
 
+    def _get_valid_due_date(self, date):
+        if isinstance(date, str):
+            date = fields.Date.from_string(date)
+        is_date_in_holiday = self.is_date_in_holiday(date)
+        while is_date_in_holiday:
+            date = is_date_in_holiday[1] + relativedelta(days=1)
+            is_date_in_holiday = self.is_date_in_holiday(date)
+        return date
+
     def is_date_in_holiday(self, date):
         if isinstance(date, str):
             date = fields.Date.from_string(date)
-        res = self.holiday_ids.filtered(
-            lambda x: int(x.month_from) <= date.month and int(x.month_to) >= date.month
-            and int(x.day_from) <= date.day and int(x.day_to) >= date.day
-        )
-        if bool(res):
-            res = res[0]
-            res_date_to = False
-            day_to = int(res.day_to)
-            while not res_date_to:
-                try:
-                    res_date_to = fields.Date.from_string(
-                        "%s-%s-%s" % (date.year, res.month_to, day_to)
-                    )
-                except:
-                    day_to -= 1
-            return [
-                fields.Date.from_string(
-                    "%s-%s-%s" % (date.year, res.month_from, res.day_from)
-                ), res_date_to
-            ]
+        for holiday in self.commercial_partner_id.holiday_ids:
+            holiday_start_date = self._generate_field_date(
+                date.year, int(holiday.month_from), int(holiday.day_from)
+            )
+            holiday_end_date = self._generate_field_date(
+                date.year, int(holiday.month_to), int(holiday.day_to)
+            )
+            if date >= holiday_start_date and date <= holiday_end_date:
+                return [holiday_start_date, holiday_end_date]
         return False
+
+    def _generate_field_date(self, year, month, day):
+        # When the user selects a date that does not exist, assume the last day
+        # for that month
+        days = (day, max(calendar.monthrange(year, month)))
+        return fields.Date.from_string("%s-%s-%s" % (year, month, min(days)))
 
 
 class ResPartnerHoliday(models.Model):
@@ -96,6 +103,14 @@ class ResPartnerHoliday(models.Model):
         string="Month to",
         required=True,
     )
+
+    _sql_constraints = [
+        (
+            'month_consistency',
+            'CHECK(month_from <= month_to)',
+            'Month from should be higher than month from'
+        ),
+    ]
 
     @api.constrains("day_from", "month_from", "day_to", "month_to")
     def _check_from_end_dates(self):
--- a/account_payment_term_partner_holiday/README.rst
+++ b/account_payment_term_partner_holiday/README.rst
@@ -44,6 +44,11 @@ To use this module, you need to:
    Invoices' and create or edit some record.
 #. If the computed due date is inside a holidays period, it's moved to the first available date.
 
+Known issues / Roadmap
+======================
+
+* Due dates for invoices are not updated when new holidays are created after.
+
 Bug Tracker
 ===========
 
--- a/account_payment_term_partner_holiday/__manifest__.py
+++ b/account_payment_term_partner_holiday/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "Account Payment Term Partner Holiday",
-    "version": "12.0.1.0.0",
+    "version": "12.0.1.1.0",
     "website": "https://github.com/OCA/account-payment",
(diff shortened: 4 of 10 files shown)


account_payment_term_partner_holiday/models/res_partner.py
```diff
@@ -21,7 +21,11 @@ def is_date_in_holiday(self, date):
             date = fields.Date.from_string(date)
         res = self.holiday_ids.filtered(
             lambda x: int(x.month_from) <= date.month and int(x.month_to) >= date.month
-            and int(x.day_from) <= date.day and int(x.day_to) >= date.day
+            and int(x.day_from) <= date.day
+            and (
+                (int(x.day_to) >= date.day and int(x.month_to) >= date.month)
+                or (int(x.day_to) <= date.day and int(x.month_to) > date.month)
+            )
```
joao-p-marques: Instead of all this, can't you do something like:
1. Take the current `year` from `date` 
1. Generate a start date for the holiday period for that year from `holiday_ids` 
3. Do the same for an end date
3. Check if `date` is between them with normal `Date` object comparison

account_payment_term_partner_holiday/models/account_invoice.py
```diff
@@ -17,7 +17,7 @@ def _onchange_payment_term_date_invoice(self):
 
     @api.onchange("date_due")
     def _onchange_date_due_account_payment_term_partner_holiday(self):
-        if self.date_due and self.payment_term_id and self.partner_id:
+        if self.date_due and self.partner_id:
```
joao-p-marques, line 19: Also, shouldn't this be applied only when there is a payment term selected? Maybe the other fix should be enough and this should be restored...

pedrobaeza: Due dates can be set without payment term, so if this condition is limiting the auto-change of the due date, then it should be removed.

victoralmau: I have removed `payment_term_id` condition because `payment_term_id` field is not required to recalculate "allowed" `due_date`.

joao-p-marques: Ok, it is just that the module name and all the docs suggest that this will only be applied when a payment term is selected (which makes sense as that is when the due date will be computed, in other cases it is manually selected, right?) Maybe we could update the docs to reflect that...

account_payment_term_partner_holiday/models/res_partner.py
joao-p-marques: Maybe something like this could work @victoralmau:

```python
def is_date_in_holiday(self, date):
    if isinstance(date, str):
        date = fields.Date.from_string(date)
    for holiday in self.holiday_ids:
        holiday_start_date = self._generate_field_date(date.year, holiday.month_from, holiday.day_from)
        # If the user selected a date of the next year (e.g. vacations from December to January), compute the correct date
        holiday_end_date = self._generate_field_date(year_to, holiday.month_to, holiday.day_to) if (holiday.month_to > holiday.month_from or (holiday.month_to == holiday_month_from and holiday.day_to > holiday.day_from)) else self._generate_field_date(date.year + 1, holiday.month_to, holiday.day_to)
    	if date >= holiday_start_date and date <= holiday_end_date:
            return [holiday_start_date,holiday_end_date]
    return False

def _generate_field_date(year, month, day):
    try:
        return fields.Date.from_string("%s-%s-%s" % (year, month, day))
    except ValueError:
        # When the user selects a date that does not exist, assume the last day for that month
        return fields.Date.from_string("%s-%s-%s" % (year if month < 12 else year + 1, month + 1 if month < 12 else 1, 1)) - timedelta(day=1)
```

victoralmau: ping @pedrobaeza what do you think about these idea to change function?

pedrobaeza: Seems good, except some code hints:

- You can compare directly a date interval: `min > date > max`
- Why substracting one day?

joao-p-marques: > Why substracting one day?

This is a trick to get the last day of the current month: Go to the first day of the following month and subtract 1.

pedrobaeza: I think you can get that with `calendar` from dateutils, but OK as well.

joao-p-marques: Indeed, I see there are a couple of ways of doing this:

1. Use `datetutil.relativedelta`. Is is similar to `timedelta` but, if you add 31 days to any day of the month, you willl get the last day of that month: https://dateutil.readthedocs.io/en/stable/relativedelta.html
2. Use `calendar.monthrange` that returns the number of days in a month, so you can get if from there: https://docs.python.org/3/library/calendar.html#calendar.monthrange

@victoralmau it is up to you if you want to change it or not. Any of the solutions look good to me 👍

joao-p-marques (approved):

account_payment_term_partner_holiday/models/res_partner.py
victoralmau: I have changed to `calendar.monthrange`
According to these change, it's a good idea to change `_generate_field_date()` too to simplify?

```
def _generate_field_date(self, year, month, day):
    # When the user selects a date that does not exist, assume the last day
    # for that month
    return fields.Date.from_string(
        "%s-%s-%s" % (year, month, max(calendar.monthrange(year, month)))
    )
```

pedrobaeza: It should be `min` and include the end day.

joao-p-marques: I would maybe keep the comment or a docstring in `_generate_field_date` to explain what you are doing, otherwise LGTM :+1:

pedrobaeza:
Waiting for the latest commented changes: commercial partner + table visibility

victoralmau:
New changes done (commercial partner + table visibility)

account_payment_term_partner_holiday/models/account_invoice.py
```diff
@@ -17,12 +17,12 @@ def _onchange_payment_term_date_invoice(self):
...
             new_date_due = self.date_due
             is_date_in_holiday = self.partner_id.is_date_in_holiday(new_date_due)
-            if is_date_in_holiday:
-                res_date_end = is_date_in_holiday[1]
-                new_date_due = res_date_end + relativedelta(days=1)
+            while is_date_in_holiday:
```
pedrobaeza: Why coming back again to a loop? This kind of loops are always dangerous.

pedrobaeza: And why not having this stuff in a method for being shared in all code? Instead of `is_date_in_holiday`, put the method `_get_valid_due_date` or similar in the partner and given a date, return the next valid date, and call it from both places.

account_payment_term_partner_holiday/models/res_partner.py
```diff
@@ -19,28 +21,28 @@ class ResPartner(models.Model):
...
-                ), res_date_to
-            ]
+        for holiday in self.commercial_partner_id.holiday_ids:
+            holiday_start_date = self._generate_field_date(
+                date.year, int(holiday.month_from), int(holiday.day_from)
+            )
+            # If the user selected a date of the next year (e.g. vacations from
+            # December to January), compute the correct date
+            year_end = date.year
+            if holiday.month_to < holiday.month_from:
```
pedrobaeza: This should be forbidden in the creation, not covered here. We already talked that the constraint should be `month_from` always < `month_to`.

pedrobaeza: Is the table ordered by day_from and month_from by the way? It should.

pedrobaeza (changes requested):

account_payment_term_partner_holiday/models/account_invoice.py
victoralmau: Loop is necessary because:
- If actually `is_date_in_holiday()` function get first holiday record (with date_from and date_to), we will use date_to + 1 day to new date_due but, what happen if new date is too unavailable according to holidays?
- If `is_date_in_holiday()` function would return last date_to from holiday is absolutely incorrect.

IMO loop is the only way to get first new date due available according holidays.
What do you think about it?
It's correct "replace"  `is_date_in_holiday` to `_get_valid_due_date` to make loop inside and return only first valid date?

pedrobaeza: OK about a loop, but not going on steps of 1 day. Just set next day as holidays end + 1, and repeat the check.

victoralmau: Ok, but we keep loop or is better create `_get_valid_due_date()` function to do it?

pedrobaeza: Create the function for not doing 2 times the same thing according my comment in https://github.com/OCA/account-payment/pull/429#discussion_r669337386

account_payment_term_partner_holiday/models/res_partner.py
```diff
@@ -16,31 +19,36 @@ class ResPartner(models.Model):
...
+    def _get_valid_due_date(self, date):
+        if isinstance(date, str):
+            date = fields.Date.from_string(date)
+        is_date_in_holiday = self.is_date_in_holiday(date)
+        while is_date_in_holiday:
+            date = is_date_in_holiday[1]
+            is_date_in_holiday = self.is_date_in_holiday(date)
+        return date
+
     def is_date_in_holiday(self, date):
```
joao-p-marques, line 31: As this is more than just a check, I would maybe rename the function (to `date_in_holiday`, for example) and add a docstring explaining that it should return the last day of vacation in case there is any for this period (check also my next comment for context)

account_payment_term_partner_holiday/models/res_partner.py
```diff
@@ -16,31 +19,36 @@ class ResPartner(models.Model):
...
+        for holiday in self.commercial_partner_id.holiday_ids:
+            holiday_start_date = self._generate_field_date(
+                date.year, int(holiday.month_from), int(holiday.day_from)
+            )
+            holiday_end_date = self._generate_field_date(
+                date.year, int(holiday.month_to), int(holiday.day_to)
+            )
+            if date >= holiday_start_date and date <= holiday_end_date:
+                holiday_end_date += relativedelta(days=1)
+                return [holiday_start_date, holiday_end_date]
```
joao-p-marques: Are there any situations where you will use the start date? If not, maybe it would be easier to return only the end date... Also, by the context of the function, IMHO it would be clearer if `_get_valid_due_date` was the method in charge of selecting the **next available date**, and this one should only return the **last day of vacation**. That is only to make it clearer, in my opinion, and is not completely necessary 
```suggestion
                return holiday_end_date
```

account_payment_term_partner_holiday/models/res_partner.py
```diff
@@ -16,31 +19,36 @@ class ResPartner(models.Model):
         auto_join=True
     )
 
+    def _get_valid_due_date(self, date):
+        if isinstance(date, str):
+            date = fields.Date.from_string(date)
+        is_date_in_holiday = self.is_date_in_holiday(date)
+        while is_date_in_holiday:
+            date = is_date_in_holiday[1]
```
joao-p-marques: Then this would change according to the comments on the other function:
```suggestion
            date = is_date_in_holiday + relativedelta(1)
```

pedrobaeza: Shouldn't it be?:
```suggestion
            date = is_date_in_holiday[1] + relativedelta(days=1)
```

joao-p-marques: As commented [above](https://github.com/OCA/account-payment/pull/429#discussion_r669620630), if we never use the start date, why do we need to return it and then have to select the specific element? This was imagining we only returned the end date in the other function...

victoralmau:
Thanks @joao-p-marques , tomorrow I will apply the changes you indicate to simplify the functions and return only one date.

pedrobaeza:
But this is not what I want, as the idea is to have the interval for any possible extension. It's not being used right now, but it may be useful on the future.

account_payment_term_partner_holiday/models/res_partner.py
joao-p-marques: > But this is not what I want, as the idea is to have the interval for any possible extension. It's not being used right now, but it may be useful on the future.

Ok, it was only a suggestion. In that case, just follow @pedrobaeza’s suggestion 👍

account_payment_term_partner_holiday/models/res_partner.py
joao-p-marques: Also, ignore this

pedrobaeza (approved):

PR #432, opened 2021-07-21
[12.0][IMP+FIX] account_payment_term_partner_holiday: Change addon to depends account_payment_term_extension 

victoralmau:
Changes:
- [IMP] Change addon to depends `account_payment_term_extension` 
- [FIX] Change the way to set correctly due date according payment term lines definition (not +1 day always).

It's necessary to not use `_get_valid_due_date()` in `compute()` payment term process to prevent incorrect dates (before if we set some date, allways return +1 day according to "last" holiday period but what happen when +10 days according to payment term line or if payment term line set custom payment days?), we only use these function when due date change manually (not if date have change according to payment term set).

PD: In 13.0 is different because we only set payment term OR due date in invoice, it's not possible to change both.

**Some examples to understand these changes.**

Partner holidays: From 2021-01-01 to 2021-01-15
Payment term A: +0 days
Payment term B: +10 days
Payment term C: custom payment date: 5 or 10
Payment term D: have holiday in 2021-01-16, date postponed is 2021-03-01

| Invoice date  | Payment term  | Before due date | Current due date |
| ------------- | ------------- | --------------- | ---------------- |
| 2021-01-01 | Payment term A | 2021-01-16 | 2021-01-16 |
| 2021-01-01 | Payment term B | 2021-01-16 | 2021-01-16 |
| 2021-01-01 | Payment term C |  | 2021-02-05 |
| 2021-01-01 | Payment term D |  | 2021-03-01 |

Invoice B: Before these change it's ok, but now is 2021-01-26.
Invoice C: Before these change expected due date was 2021-01-16 but it's incorrect, we can't re-create all possibilities in override function AND it's reason to allway re-compute only when new due date isn't initial due date.
Invoice D: The only way to be sure that generated due date is correct is re-compute at end.

@Tecnativa TT31110

### DIFF
--- a/account_payment_term_partner_holiday/models/account_invoice.py
+++ b/account_payment_term_partner_holiday/models/account_invoice.py
@@ -16,7 +16,15 @@ class AccountInvoice(models.Model):
 
     @api.onchange("date_due")
     def _onchange_date_due_account_payment_term_partner_holiday(self):
-        if self.date_due and self.partner_id:
+        """Recompute the due date to the next available date according to
+        the holiday periods set in the partner.
+
+        It must only be re-calculated when a payment term is not set.
+        This prevents the due date to be changed again and that another
+        given number of days are added according to what is set on the
+        payment term.
+        """
+        if self.date_due and self.partner_id and not self.payment_term_id:
             new_date_due = self.partner_id._get_valid_due_date(self.date_due)
             if new_date_due != self.date_due:
                 self.date_due = new_date_due
--- a/account_payment_term_partner_holiday/models/account_payment_term.py
+++ b/account_payment_term_partner_holiday/models/account_payment_term.py
@@ -9,16 +9,27 @@ class AccountPaymentTerm(models.Model):
 
     @api.one
     def compute(self, value, date_ref=False):
+        """Compute the due date taking into account the holiday periods
+        set in the partner.
+
+        Once an initial date resulting of the payment term is computed,
+        compute the first available date after that.
+        Then, apply_payment_days() and apply_holidays() to prevent
+        incompatibilities.
+        """
         result = super().compute(value=value, date_ref=date_ref)[0]
         ctx = self.env.context
         partner_id = ctx.get("move_partner_id", ctx.get("default_partner_id"))
         if partner_id:
             partner = self.env["res.partner"].browse(partner_id)
             result2 = []
-            for item in result:
-                new_date_item = partner._get_valid_due_date(item[0])
-                if new_date_item != item[0]:
-                    result2.append((fields.Date.to_string(new_date_item), item[1]))
+            for key, item in enumerate(result):
+                next_date = partner._get_valid_due_date(item[0])
+                if next_date != item[0]:
+                    line = self.line_ids[key]
+                    next_date = self.apply_payment_days(line, next_date)
+                    next_date = self.apply_holidays(next_date)
+                    result2.append((fields.Date.to_string(next_date), item[1]))
                 else:
                     result2.append(item)
             result = result2
--- a/account_payment_term_partner_holiday/tests/test_partner_holiday.py
+++ b/account_payment_term_partner_holiday/tests/test_partner_holiday.py
@@ -57,6 +57,27 @@ class TestPartnerHoliday(common.TransactionCase):
                 "option": "day_after_invoice_date"
             })]
         })
+        self.payment_term_with_holidays = self.env["account.payment.term"].create({
+            "name": "Immediate (with holidays)",
+            "line_ids": [(0, 0, {
+                "value": "balance",
+                "days": 0,
+                "option": "day_after_invoice_date"
+            })],
+            "holiday_ids": [(0, 0, {
+                "holiday": "2021-06-14",
+                "date_postponed": "2021-07-08"
+            })]
+        })
+        self.payment_term_immediate_custom = self.env["account.payment.term"].create({
+            "name": "Immediate (with custom payment days)",
+            "line_ids": [(0, 0, {
+                "value": "balance",
+                "days": 0,
+                "option": "day_after_invoice_date",
+                "payment_days": "5,10"
+            })]
+        })
         self.journal = self.env["account.journal"].create({
             "name": "Test sale",
             "type": "sale",
@@ -154,6 +175,14 @@ class TestPartnerHoliday(common.TransactionCase):
         invoice_form.payment_term_id = self.payment_term_10_days
         self.assertEqual(invoice_form.date_due, fields.Date.from_string("2021-04-11"))
 
+    def test_invoice_payment_term_custom_partner_1(self):
+        invoice_form = self._set_invoice_form(self.partner_1.id, "2021-06-12")
+        invoice_form.payment_term_id = self.payment_term_with_holidays
+        self.assertEqual(invoice_form.date_due, fields.Date.from_string("2021-07-08"))
+        invoice_form = self._set_invoice_form(self.partner_1.id, "2021-02-01")
+        invoice_form.payment_term_id = self.payment_term_immediate_custom
+        self.assertEqual(invoice_form.date_due, fields.Date.from_string("2021-04-05"))
+
     def test_invoice_payment_term_partner_1_child(self):
         invoice_form = self._set_invoice_form(self.partner_1_child.id, "2021-02-01")
         invoice_form.payment_term_id = self.payment_term_immediate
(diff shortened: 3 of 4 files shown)


account_payment_term_partner_holiday/models/account_invoice.py
```diff
@@ -16,7 +16,11 @@ def _onchange_payment_term_date_invoice(self):
 
     @api.onchange("date_due")
     def _onchange_date_due_account_payment_term_partner_holiday(self):
-        if self.date_due and self.partner_id:
+        """Is necessary to re-calculate only when payment_term_id is not set
+        to prevent date_due is changed again and apply +10 days for example
+        according to payment term.
+        """
```
joao-p-marques: If you are adding a docstring, it would be better to explain the full behavior of the method. I think this would make it clearer:

```suggestion
        """Recompute the due date to the next available date according to
        the holiday periods set in the partner.
        
        It must only be re-calculated when a payment term is not set.
        This prevents the due date to be changed again and that another 
        given number of days are added according to what is set on the 
        payment term.
        """
```

Or something like that, if you think it is not completely right...

account_payment_term_partner_holiday/models/account_payment_term.py
```diff
@@ -2,24 +2,43 @@
...
+                    )[0]
+                    res_is_date_in_holiday = partner.is_date_in_holiday(res[0][0])
+                # Need to re-compute according to initial comment
+                if date_to_check != item[0]:
+                    res = super().compute(
+                        value=value, date_ref=date_to_check
+                    )[0]
+                    date_to_check = res[0][0]
+                new_item = (date_to_check, item[1])
+                result2.append(new_item)
```
joao-p-marques: If this is the actual desired behavior (always recompute the date), I think this can be simplified, following the current logic:

```suggestion
                new_date_item = partner._get_valid_due_date(item[0])
                if new_date_item != item[0]:
                    # Re-compute
                    res = super().compute(
                        value=value, date_ref=new_date_item
                    )[0]
                    result2.append(res[0])
                else:
                    result2.append(item)
            result = result2
```

This way, you avoid:

1. Repeating code that you already have in `res.partner` for the incremental date check
2. Re-compute the payment term result date twice (as for your explanation, you just need to recompute once when you already have the first available date computed)

account_payment_term_partner_holiday/models/account_payment_term.py
```diff
@@ -2,24 +2,43 @@
...
 class AccountPaymentTerm(models.Model):
     _inherit = "account.payment.term"
 
     @api.one
     def compute(self, value, date_ref=False):
+        """If date_to_check has changed is because initial date is in holiday
+        period and these change make necessary to re-compute again to be sure
+        that date returnes is according to payment term (+10 days or custom
+        payment days).
+        """
```
joao-p-marques: As with before, maybe a more complete explanation would be clearer here:
```suggestion
        """Compute the due date taking into account the holiday periods
        set in the partner.
         
        Once an initial date resulting of the payment term is computed, compute the first 
        available date after that.
        Then, recompute the due date applying the payment term to the 
        previously computed date.
        """
```

joao-p-marques (changes requested):
Code and functional review.

Having to recompute the due date when there is a payment term set at the end is a bit weird and can cause some cases that might be hard to understand functionally. However, I am not sure how it could be done better and maybe this is the best option. Maybe @pedrobaeza can give some input here.

Anyway, if this is the approach to follow, please check the comments to improve performance and overall make the code easier to read.

account_payment_term_partner_holiday/tests/test_partner_holiday.py
```diff
@@ -55,6 +55,19 @@ def setUp(self):
                 "value": "balance",
                 "days": 10,
                 "option": "day_after_invoice_date"
+            })],
+            "holiday_ids": [(0, 0, {
```
pedrobaeza, line 67: Can you separate tests for payment day for not changing existing tests?

account_payment_term_partner_holiday/tests/test_partner_holiday.py
```diff
@@ -140,45 +161,72 @@ def test_invoice_payment_term_partner_1(self):
         invoice_form.payment_term_id = self.payment_term_immediate
         self.assertEqual(invoice_form.date_due, fields.Date.from_string("2021-04-02"))
         invoice_form.payment_term_id = self.payment_term_10_days
-        self.assertEqual(invoice_form.date_due, fields.Date.from_string("2021-04-02"))
+        self.assertEqual(invoice_form.date_due, fields.Date.from_string("2021-04-12"))
```
pedrobaeza: You are still modifying previous tests

victoralmau: Sorry but i haven't modify tests.
Before these changes, due date generated is not correctly because not re-compute dates (only returned +1 days according to "last" holiday period), now final result in use cases is different.

pedrobaeza: So you are telling me that there was an error in previous tests? If so, you should split the commit with the corresponding changes and explain the issue.

joao-p-marques: I think what @victoralmau means is that this PR changes the behavior (that's why I asked [here](https://github.com/OCA/account-payment/pull/432#pullrequestreview-712391900) if this is the best approach) and, as a result, current tests need to be changed...

pedrobaeza: I have a short-circuit about this, as I don't conceive that the dates vary this way due to that change. Is it only on tests or it's something that is happening on the behavior of the module? If it's the first, we need to create more resilient tests. If it's the second, that behavior change should be analyzed.

joao-p-marques: It is the second, and I agree that the behavior change should be analyzed, that is why I pinged you about it above:

> Having to recompute the due date when there is a payment term set at the end is a bit weird and can cause some cases that might be hard to understand functionally. However, I am not sure how it could be done better and maybe this is the best option. Maybe @pedrobaeza can give some input here.

This basically means that, instead of selecting the first available date, the module now selects the first available one that matches the payment term "restrictions" (last day of the month, 5th day of the month, etc.)

As I said, I am not sure this suggestion from @victoralmau  is the best way of doing this, as it has a pretty big functional impact, so you will say what you prefer for it...

pedrobaeza: Well, we are adding here an improvement for taking into account payment days, but that shouldn't impact on the previous tests suite and data, that didn't have such data to be altered, so I still don't understand why the change.

victoralmau: I have modified the description with some examples to try to explain some use cases.

Yes, some use cases now have change expected due date (A and B) but IMO is the only way to get allways valid due date (C and D).

pedrobaeza:
Reading your main comment, I'm not sure if "Current due date" means the due date got with this PR, but it so then
```
Invoice B: Before these change it's ok, but now is 2021-01-26.
```
It's incorrect. It should be 2021-01-16.

About `2021-01-01 | Payment term C |   | 2021-01-05`, the correct one should `2021-02-05`.

joao-p-marques:
About the current due date for `Payment Term C`, I think @victoralmau meant `2021-02-05`. At least, from the code logic, I understand that will be the behavior.

OTOH, to obtain what @pedrobaeza is asking, you probably need to check the Payment Term type and only add the re-computation for certain ones (in this case, for PTs C and D, once the above issue is solved), i. e., for the ones that should match a specific day of the month.

victoralmau:
Maybe a good solution to prevent side effects in Invoice B is:
- Improve `compute()` function according to payment term in `account_payment_term_extension` addon to return too payment term line.
 - Improve `compute()` function according to `account_payment_term_partner_holiday` addon and check first super without date change and after, according to payment term line check if is really necessary a recompute extra (Invoice C and Invoice D).

What do you think about it? Can you think of another solution to know when to apply that extra recalculation in some cases?

victoralmau:
Changes done and all use cases is ok now, right?

pedrobaeza (approved):

joao-p-marques:
I am afraid this still doesn't work as expected :confused:. In the following case, from local testing with this version of the code, the partner has holidays from `08/01` to `08/31`:

image

From what was mentioned earlier, I understand the correct due date in this case should be `09/30`, to match the end of the month, right?

pedrobaeza:
Such case is indeed not included, but I'm not caring very much about it.

pedrobaeza:
Please move it to v13.

PR #439, opened 2021-08-10
[14.0][MIG] account_check_date

ps-tubtim:
Standard Migration.


account_check_date/views/account_payment_views.xml
```diff
@@ -0,0 +1,29 @@
...
+        <field name="model">account.payment</field>
+        <field
+            name="inherit_id"
+            ref="account_check_printing.view_account_payment_form_inherited"
+        />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='check_number']" position="before">
+                <field
+                    name="check_date"
+                    attrs="{'invisible': ['|', ('payment_method_code', '!=', 'check_printing'), ('check_number', '=', False)]}"
```
kittiu:
```suggestion
                    attrs="{'invisible': [('payment_method_code', '!=', 'check_printing')]}"
```

kittiu: I think check_date shouldn't be invisible when check_number = False.
In case user create Payment directly, now, where is no way to enter check_date during state draft (and so, we can't enter check_date at all, even we have check_number as state is not draft)

WDYT?

Selection_164

kittiu (approved):

Saran440 (approved):
Functional Test :+1:

PR #440, opened 2021-08-10
[IMP] Documentation (dummy commit to fix the missing package on Pypi)

max3903:
For some reason, the package is not in https://pypi.org.


bodedra (approved):
LG :+1:

PR #444, opened 2021-08-12
[13.0][IMP+FIX] account_payment_term_partner_holiday: Change addon to depends account_payment_term_extension

victoralmau:
FWP from 12.0: https://github.com/OCA/account-payment/pull/432

Changes:
- [IMP] Change addon to depends `account_payment_term_extension` 
- [FIX] Change the way to set correctly due date according payment term lines definition (not +1 day always).

It's necessary to not use `_get_valid_due_date()` in `compute()` payment term process to prevent incorrect dates (before if we set some date, allways return +1 day according to "last" holiday period but what happen when +10 days according to payment term line or if payment term line set custom payment days?), we only use these function when due date change manually (not if date have change according to payment term set).

PD: In 13.0 is different because we only set payment term OR due date in invoice, it's not possible to change both.

**Some examples to understand these changes.**

Partner holidays: From 2021-01-01 to 2021-01-15
Payment term A: +0 days
Payment term B: +10 days
Payment term C: custom payment date: 5 or 10
Payment term D: have holiday in 2021-01-16, date postponed is 2021-03-01

| Invoice date  | Payment term  | Before due date | Current due date |
| ------------- | ------------- | --------------- | ---------------- |
| 2021-01-01 | Payment term A | 2021-01-16 | 2021-01-16 |
| 2021-01-01 | Payment term B | 2021-01-16 | 2021-01-16 |
| 2021-01-01 | Payment term C |  | 2021-02-05 |
| 2021-01-01 | Payment term D |  | 2021-03-01 |

Invoice B: Before these change it's ok, but now is 2021-01-26.
Invoice C: Before these change expected due date was 2021-01-16 but it's incorrect, we can't re-create all possibilities in override function AND it's reason to allway re-compute only when new due date isn't initial due date.
Invoice D: The only way to be sure that generated due date is correct is re-compute at end.

@Tecnativa TT31110

### DIFF
--- a/account_payment_term_partner_holiday/__manifest__.py
+++ b/account_payment_term_partner_holiday/__manifest__.py
@@ -9,7 +9,7 @@
     "license": "AGPL-3",
     "application": False,
     "installable": True,
-    "depends": ["account"],
+    "depends": ["account_payment_term_extension"],
     "maintainers": ["victoralmau"],
     "development_status": "Production/Stable",
     "data": ["security/ir.model.access.csv", "views/res_partner_view.xml"],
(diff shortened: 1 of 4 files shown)


pedrobaeza (approved):
Direct fw-port, so fast-tracking it

PR #443, opened 2021-08-11
[14.0][MIG] account_payment_return_import_iso20022

pablo-lp:

PR #445, opened 2021-08-25
[MIG] account_payment_return_import_iso20022: Migration to 14.0

jaiodoo:


pedrobaeza:
You are not preserving commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0.

But there's also #443 for migrating the modules, so closing. Please review that one instead.

PR #449, opened 2021-09-21
[14.0][ADD] account_payment_cancel_confirm

ps-tubtim:
Depends on,

- [x] https://github.com/OCA/server-ux/pull/297

This module extend `base_cancel_confirm` to make cancel confirmation with cancel reason (none / optional / requried).


kittiu (approved):
LGTM

Saran440:
We can't split module payment and move because payment cancel depend on cancel journal entries and it need reason cancel.

I think this PR should be move on https://github.com/OCA/account-invoicing/pull/872

kittiu:
Yes, I think this is not needed. payment and invoice are sharing object in v14.
@ps-tubtim

ps-tubtim:
Move to https://github.com/OCA/account-invoicing/pull/872

PR #455, opened 2021-09-29
[13.0][FIX] account_payment_promissory_note: Due Date not moved to account moves

CarlosRoca13:
The method `_get_liquidity_move_line_vals` does not exist on v13, so we change it to new method to extend the due date to the related account moves.

cc @Tecnativa TT32207

please @sergio-teruel @carlosdauden review it


pedrobaeza:
Can you add a test for avoiding further regressions?

CarlosRoca13:
Added @pedrobaeza

pedrobaeza (approved):

sergio-teruel (approved):

PR #457, opened 2021-10-13
12.0 add payment slimpay

fcayre:


legalsylvain:
thanks for porting this module.

partial review : 
- you should add dependency to ``coreapi`` python lib
- you should reference the repo OCA/base_phone in the oca_dependencies.txt file. (because of base_phone dependencies)
- also some flake8 / pylint errors.

fcayre:
@legalsylvain coreapi is a dependency of hal_codec (see https://github.com/core-api/python-hal-codec/blob/master/requirements.txt) which is present in the manifest, so I think it is OK like this.
Thanks for the review, just pushed python style fixes, let's see how the CI goes...

PR #465, opened 2021-11-16
[14.0][IMP] account_due_list: adding remaining days widget

chusamo:
Captura de pantalla de 2021-11-16 08-40-59


angelmoya:
Hi, works fine, but... our financial users usually like to see date better than see remaining days... maybe you can add twice as optional show... this way user can decide to use they like

dreispt (approved):

PR #467, opened 2021-12-08
[14.0][ENH] account_payment_multi_deduct: edit register payment view

ps-tubtim:
This pull request is used to edit register payment view because in the case that there is an additional field it will make filling in information easier.

**Before:**
Selection_176

**After:** 
Selection_177


kittiu:
Should be [ENH]

kittiu (approved):

Saran440 (approved):

PR #472, opened 2021-12-17
[13.0][FIX] account_due_list_days_overdue: Call postprocess method for simpler code

victoralmau:
Change the order of the parameters for the `transfer_modifiers_to_node()` function according to Odoo.

https://github.com/odoo/odoo/blob/e916526adce0f10f3d22badd43f59ac405ea96c3/odoo/addons/base/models/ir_ui_view.py#L94-L97
```
def transfer_modifiers_to_node(modifiers, node):
    if modifiers:
        simplify_modifiers(modifiers)
        node.set('modifiers', json.dumps(modifiers))
```

@Tecnativa

### DIFF
--- a/account_due_list_days_overdue/models/account_move_line.py
+++ b/account_due_list_days_overdue/models/account_move_line.py
@@ -8,11 +8,6 @@ from lxml import etree
 
 from odoo import api, fields, models
 
-from odoo.addons.base.models.ir_ui_view import (
-    transfer_modifiers_to_node,
-    transfer_node_to_modifiers,
-)
-
 
 class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Link to the Odoo code about that. I think there are other methods at higher abstraction layer.

joao-p-marques (approved):
Looks good 👍 

@victoralmau can you check @pedrobaeza’s comment on possible methods higher in the abstraction layer?

victoralmau:
> Looks good +1
> 
> @victoralmau can you check @pedrobaeza’s comment on possible methods higher in the abstraction layer?

Checking Odoo code again, in my opinion there is no better method we can use.

pedrobaeza:
I was asking for the link. Looking at the `fields_view_get` method, this is the one I'm referring to:

https://github.com/odoo/odoo/blob/7ee5cd3df28d9ad9337e5f231b6c855a3c4f329c/odoo/models.py#L1564

Can you please check to call this one? Or there is something I'm missing?

victoralmau:
> I was asking for the link. Looking at the `fields_view_get` method, this is the one I'm referring to:
> 
> https://github.com/odoo/odoo/blob/7ee5cd3df28d9ad9337e5f231b6c855a3c4f329c/odoo/models.py#L1564
> 
> Can you please check to call this one? Or there is something I'm missing?

I have changed the code to use this function (`postprocess_and_fields`). I have also fixed the sum of the values in the tree view (the sum was not displayed).

pedrobaeza (approved):

PR #476, opened 2022-01-11
[FIX] account_cash_invoice: wizards and account_bank_statement model

olgamarcocb:
As can be seen in the attached videos, this PR fixes the "Reconciled Entries" button as well as the statement lines:

Previously:
Before
Applying changes:
After

- The field "counterpart_account_id" is added on cash_invoice_in and cash_invoice_out wizards.
- The function "button_post" is added on account_bank_statement model.

### DIFF
--- a/account_cash_invoice/models/account_bank_statement.py
+++ b/account_cash_invoice/models/account_bank_statement.py
@@ -1,0 +1,20 @@
+# Copyright 2021 Creu Blanca
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class AccountBankStatement(models.Model):
+
+    _inherit = "account.bank.statement"
+
+    def button_post(self):
+        lines_of_moves_to_reconcile = self.line_ids.filtered(
+            lambda line: line.move_id.state != "posted" and line.invoice_id
+        )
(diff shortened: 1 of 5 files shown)


etobella (approved):
Thanks! Tested locally

account_cash_invoice/models/account_bank_statement.py
```diff
@@ -0,0 +1,19 @@
...
+
+class AccountBankStatement(models.Model):
+
+    _inherit = "account.bank.statement"
+
+    def button_post(self):
+        lines_of_moves_to_reconcile = self.line_ids.filtered(
+            lambda line: line.move_id.state != "posted" and line.invoice_id
+        )
+        super(AccountBankStatement, self).button_post()
```
acsonefho: Minor change:
You should save the result (even if it's `None`) and return it at the end.
```
result = super().button_post()
...
return result
```

acsonefho (commented):
Minor comment

account_cash_invoice/models/account_bank_statement.py
olgamarcocb: Thank you! Done :smile:

AaronHForgeFlow (approved):
Functional test :+1:

PR #477, opened 2022-01-16
[14.0][ADD] account_payment_multi_deduction_analytic

Saran440:
This module add analytic account and tag on register payment (multi deduction) for send analytic and tag into journal items.
Selection_004
- it default if you register payment with same analytic.

and it will send analytic to journal items
Selection_005


Saran440:
move on https://github.com/OCA/account-payment/pull/478

PR #479, opened 2022-01-24
[14.0][FIX] account_payment_return: use journal return account

ao-landoo:
This fix allows to reconcile the journal entry created from a payment.return with the selected bank. Without the fix the counterpart used in the journal entry is the default_account_id, the fix replaces this account with the payment_credit_account_id which is reconcilable with the bank statement line.

### DIFF
--- a/account_payment_return/models/payment_return.py
+++ b/account_payment_return/models/payment_return.py
@@ -176,7 +176,7 @@ class PaymentReturn(models.Model):
             "name": move.ref,
             "debit": 0.0,
             "credit": total_amount,
-            "account_id": self.journal_id.default_account_id.id,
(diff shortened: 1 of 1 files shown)


rp-landoo (approved):
Functional approval

ao-landoo:
@joao-p-marques Could you review this PR as you migrated this module to V14? Thanks in advance

ao-landoo:
@ernestotejeda @rousseldenis @carlosdauden  Could you review this?

rousseldenis:
@ao-landoo Could you add a test that highlight this change ?

carlosdauden (approved):

PR #464, opened 2021-11-10
[14.0][MIG] account_financial_discount

grindtildeath:
Migrated from https://github.com/OCA/account-financial-tools/pull/1040

PR #483, opened 2022-02-01
[14.0][MIG] account_financial_discount

i-vyshnevska:
moved https://github.com/OCA/account-payment/pull/464 to finish and fix tests


account_financial_discount/tests/test_financial_discount_reconciliation.py
```diff
@@ -425,38 +426,39 @@ def test_vendor_bill_eur_with_tax_bank_reconciliation(self):
...
-            write_off_vals[1].get("credit"),
-            self.eur_currency._convert(
-                self.amount_discount_tax,
-                self.usd_currency,
-                vendor_bill.company_id,
-                vendor_bill.invoice_date,
-            ),
-        )
+        self.assertFalse(matching_amls.get(st_line.id).get("status"))
+        # reconciliation for different currency declined
```
simahawk, line 430: do we need this?

simahawk (commented):
@i-vyshnevska can you squash some commits?

@TDu could you add a review?

account_financial_discount/models/account_move.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    has_discount_available = fields.Boolean(
+        "Has discount available",
+        compute="_compute_financial_discount_data",
+        search="_search_has_financial_discount",
+        # TODO implement read_group?
```
TDu, line 17: I don't understand this todo

account_financial_discount/models/account_move.py
```diff
@@ -0,0 +1,231 @@
...
+        # Inspired by mrp.production _search_is_planned
+        if operator not in ("=", "!="):
+            raise UserError(_("Invalid domain operator %s", operator))
+        if value not in (False, True):
+            raise UserError(_("Invalid domain right operand %s", value))
+        operator_funcs = {"=": equals, "!=": not_equals}
+        move_ids = [
+            move.id
+            for move in self.search([])
+            if operator_funcs[operator](value, move.has_discount_available)
```
mmequignon, line 136: :fire: MemoryError.
We noticed this with FCI, I'm on it.

account_financial_discount/tests/test_financial_discount_reconciliation.py
i-vyshnevska: possibility to reconcile different currencies is mentioned on todo list and it is disabled because the issue in odoo
probably we may restore this later

account_financial_discount/models/account_move.py
```diff
@@ -0,0 +1,231 @@
...
+    def _search_has_financial_discount(self, operator, value):
+        # Inspired by mrp.production _search_is_planned
+        if operator not in ("=", "!="):
+            raise UserError(_("Invalid domain operator %s", operator))
+        if value not in (False, True):
+            raise UserError(_("Invalid domain right operand %s", value))
+        operator_funcs = {"=": equals, "!=": not_equals}
+        move_ids = [
+            move.id
+            for move in self.search([])
```
TDu, line 135: Could we have a performance issue with this ?
Not that I would know how to improve it.

mmequignon: it tries to compute `has_discount_available` on all records at once :boom:

simahawk:
getting back to #464

PR #492, opened 2022-03-11
[14.0][FIX] Call onchange in create because invoice creation otherwise skips it

fkantelberg:
If you specify a discount for a purchase.order the payment term will be copied to the account.move. This is bypassing the onchange trigger to calculate the discount values and days on the invoice. The values aren't calculated without manually switching the payment term on the invoice.

### DIFF
--- a/account_cash_discount_base/models/account_move.py
+++ b/account_cash_discount_base/models/account_move.py
@@ -246,5 +246,13 @@ class AccountMove(models.Model):
     @api.model
     def create(self, values):
         rec = super().create(values)
-        rec._onchange_payment_term_discount_options()
+        payment_term = rec.invoice_payment_term_id
+        if not payment_term or self.move_type not in DISCOUNT_ALLOWED_TYPES:
+            return rec
+
+        if not rec.discount_percent:
+            rec.discount_percent = payment_term.discount_percent
+
+        if not rec.discount_delay:
+            rec.discount_delay = payment_term.discount_delay
         return rec
--- a/account_cash_discount_base/models/account_move.py
+++ b/account_cash_discount_base/models/account_move.py
@@ -247,12 +247,10 @@ class AccountMove(models.Model):
     def create(self, values):
         rec = super().create(values)
         payment_term = rec.invoice_payment_term_id
-        if not payment_term or self.move_type not in DISCOUNT_ALLOWED_TYPES:
+        if not payment_term or rec.move_type not in DISCOUNT_ALLOWED_TYPES:
             return rec
 
-        if not rec.discount_percent:
-            rec.discount_percent = payment_term.discount_percent
-
-        if not rec.discount_delay:
-            rec.discount_delay = payment_term.discount_delay
+        for field in ("discount_percent", "discount_delay"):
+            if field not in values:
+                rec[field] = payment_term[field]
         return rec
--- a/account_cash_discount_base/models/account_move.py
+++ b/account_cash_discount_base/models/account_move.py
@@ -242,3 +242,9 @@ class AccountMove(models.Model):
             payment_term = partner.property_supplier_payment_term_id
             res["invoice_payment_term_id"] = payment_term.id
         return res
+
+    @api.model
+    def create(self, values):
+        rec = super().create(values)
+        rec._onchange_payment_term_discount_options()
+        return rec


flotho:
many tests are failing but LGTM.
Let me know when you'll be ready for a code review.

fkantelberg:
Hello @flotho 
In my local tests it works fine and all the errors are caused in other modules and I don't see that any of it might be caused by the change.

acsonefho (changes requested):
IMO it's not correct to do it with the `create(...)`.

Maybe you can check this: https://github.com/OCA/account-payment/pull/522

fkantelberg:
@acsonefho Wouldn't it be the same? The problem here is that the `api.onchange` function isn't triggered when creating it through backend operations. If I drop the `create(...)` and call the onchange function from another onchange function I wouldn't expect a difference.

acsonefho:
> @acsonefho Wouldn't it be the same? The problem here is that the `api.onchange` function isn't triggered when creating it through backend operations. If I drop the `create(...)` and call the onchange function from another onchange function I wouldn't expect a difference.

Yes but with your code, if a user force the `discount_delay` to `10` manually (instead of `15` sets on the payment term, for example), you'll erase the custom value.

I think you should apply your `onchange` only if you are in backend operation.

fkantelberg:
> Yes but with your code, if a user force the `discount_delay` to `10` manually (instead of `15` sets on the payment term, for example), you'll erase the custom value.
> 
> I think you should apply your `onchange` only if you are in backend operation.

I understand the concerns but I can't say which operation is done by backend because it's varying a lot. The above example was a purchase.order from which the payment terms are copied. The user isn't directly creating the invoice via the frontend and no onchange will trigger causing discount_delay to be empty despite the payment term.

But yes it gets weird if an user creates an invoice manually and sets the payment term before actually creating it. Maybe we need an extra step after the onchange to ensure the values of the users given to the create function or I skip the onchange there and copy from the payment term if not set.

acsonefho:
> > Yes but with your code, if a user force the `discount_delay` to `10` manually (instead of `15` sets on the payment term, for example), you'll erase the custom value.
> > I think you should apply your `onchange` only if you are in backend operation.
> 
> I understand the concerns but I can't say which operation is done by backend because it's varying a lot. The above example was a purchase.order from which the payment terms are copied. The user isn't directly creating the invoice via the frontend and no onchange will trigger causing discount_delay to be empty despite the payment term.
> 
> But yes it gets weird if an user creates an invoice manually and sets the payment term before actually creating it. Maybe we need an extra step after the onchange to ensure the values of the users given to the create function or I skip the onchange there and copy from the payment term if not set.

I think we have the same issue. I create another PR ( #522 ) to fix that. Expect that in my case, it's by Odoo front-end.

But your fix is applied on every invoices (even for IN/OUT/Refund invoices). I don't think someone will validate your PR if you force some values (set manually by front) :/

fkantelberg:
I pushed a commit which should resolve your concerns.

acsonefho (changes requested):
Should be something like:
```
    @api.model
    def create(self, values):
        should_update_payment_terms = not any(k in values for k in ('discount_percent', 'discount_delay', 'invoice_payment_term_id'))
        rec = super().create(values)
        if should_update_payment_terms and rec.move_type not in DISCOUNT_ALLOWED_TYPES:
            # You can call your onchange here in this case
            pass
```

account_cash_discount_base/models/account_move.py
```diff
@@ -242,3 +242,17 @@ def _reverse_move_vals(self, default_values, cancel=True):
             payment_term = partner.property_supplier_payment_term_id
             res["invoice_payment_term_id"] = payment_term.id
         return res
+
+    @api.model
+    def create(self, values):
+        rec = super().create(values)
+        payment_term = rec.invoice_payment_term_id
+        if not payment_term or self.move_type not in DISCOUNT_ALLOWED_TYPES:
```
acsonefho: `self` is empty here. Should be `rec`. That won't work. But check my proposition.

fkantelberg: Does your proposition mean that you don't care about partial values? e.g. if you enter either discount_percent or discount_delay Odoo shouldn't copy from the payment term?

account_cash_discount_base/models/account_move.py
```diff
@@ -242,3 +242,15 @@ def _reverse_move_vals(self, default_values, cancel=True):
...
+    @api.model
+    def create(self, values):
+        rec = super().create(values)
+        payment_term = rec.invoice_payment_term_id
+        if not payment_term or rec.move_type not in DISCOUNT_ALLOWED_TYPES:
+            return rec
+
+        for field in ("discount_percent", "discount_delay"):
+            if field not in values:
+                rec[field] = payment_term[field]
```
acsonefho, line 255: This change will trigger the `INSERT` then an `UPDATE` (about DB).
I think it's possible to do it during the `INSERT` (by updating directly the `values` before calling `super` (only if these fields are not set manually of course)).

acsonefho (changes requested):

account_cash_discount_base/models/account_move.py
fkantelberg: I think it's hard to get the right payment_term before the super. If the payment term isn't in the values it could be a default from the field or ir.default etc...

acsonefho: Something like this (please fill it):
```
move_type = values.get('move_type')
payment_term_id = values.get('invoice_payment_term_id')
payment_term = self.env...
for field in ("discount_percent", "discount_delay"):
    if field not in values:
        values.update({field: payment_term[field]})
```

fkantelberg: Yes and what happens if payment_term_id is None? Which one do you choose? It's not as simple to just assume that we can skip it because default values in Odoo must be respected but these can come from very different places.

acsonefho: You can adapt the `if` part.
`if payment_term and field not in values:`

fkantelberg: My problem here is not that I'm unable to handle a simple None.

It's not possible to handle the case where no payment term is inside of the values dictionary. In that case Odoo fills the payment term using default values from various possible sources.

Following scenario:

1. I create an account.move without given payment_term_id as value
2. Odoo creates an account.move and uses a defined default for the payment term
3. With your `if` I would skip setting the `discount_*` values.

I agree that normally there is no default value for `invoice_payment_term_id` BUT it's possible to set one via frontend and additional modules hence the module here would skip the copying again.

PR #501, opened 2022-04-23
[FIX] account_due_list: Do not require administrator rights

ramiadavid:
We avoid having to give administrator rights to access Accounting/Payments and due list because the module needs to access the model ir.model_data


pedrobaeza (approved):

pedrobaeza:
Fast-tracking the obvious patch:

/ocabot merge patch

PR #505, opened 2022-05-16
[14.0][MIG] account due list days overdue: Migration to 14.0

victoralmau:
Migration to 14.0

Please @CarlosRoca13 and @chienandalu can you review it?

@Tecnativa

PR #506, opened 2022-05-18
[14.0][MIG] account_due_list_days_overdue: Migration to 14.0

miquelalzanillas:
Migration to 14.0


HaraldPanten (changes requested):
@miquelalzanillas You should check https://github.com/OCA/account-payment/pull/505 instead.

THX

PR #514, opened 2022-06-08
[14.0][FIX] fixed multicurrency scenario

ChrisOForgeFlow:
@ForgeFlow

### DIFF
--- a/account_payment_term_discount/wizard/account_payment_register.py
+++ b/account_payment_term_discount/wizard/account_payment_register.py
@@ -23,7 +23,7 @@ class AccountPaymentRegister(models.TransientModel):
             res.update({"invoice_id": record.id, "discount_amt": record.discount_amt})
         return res
 
-    @api.onchange("amount", "payment_difference", "payment_date")
+    @api.onchange("amount", "payment_difference", "payment_date", "currency_id")
     def onchange_payment_amount(self):
         if (
             self.invoice_id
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (commented):
Why not using amount_residual_currency?

ChrisOForgeFlow:
> 

This field exists on account_move_line but in this case needed header value, because that conversion is performed again

AaronHForgeFlow (approved):
Yes, my bad, I though it was also in the account.move.

Code review LGTM! :+1:

account_payment_term_discount/wizard/account_payment_register.py
```diff
@@ -44,7 +44,14 @@ def onchange_payment_amount(self):
                 )
                 payment_date = fields.Date.from_string(self.payment_date)
                 discount_amt = self.invoice_id.discount_amt
-
+                amount_residual = self.invoice_id.amount_residual
+                if self.invoice_id.currency_id.id != self.currency_id.id:
+                    discount_amt = self.invoice_id.currency_id.compute(
```
MiquelRForgeFlow: warning: The `compute` method is deprecated. Use `_convert` instead

the same for the other.

MiquelRForgeFlow (commented):
also add another commit to delete the useless few `from_string` from lines above.

account_payment_term_discount/wizard/account_payment_register.py
ChrisOForgeFlow: Thanks!

account_payment_term_discount/wizard/account_payment_register.py
```diff
@@ -44,7 +44,20 @@ def onchange_payment_amount(self):
                 )
                 payment_date = fields.Date.from_string(self.payment_date)
```
MiquelRForgeFlow:
```suggestion
                payment_date = self.payment_date
```

MiquelRForgeFlow: same in line 41

ChrisOForgeFlow: Done!

MiquelRForgeFlow (approved):

PR #520, opened 2022-07-09
[FIX] account_payment_return: remove wrong char

cubells:
Typo


pedrobaeza (approved):
Indeed. Not sure why this hasn't jumped before

pedrobaeza:
Fast-tracking this:

/ocabot merge patch

PR #522, opened 2022-07-14
[13.0][ADD] account_cash_discount_base_purchase

acsonefho:
Module used to auto-fill payment term (+ related fields as `discount_delay` on `account.move`) on invoice when created from purchase order.

New (glue) module because the current module `account_cash_discount_base` doesn't have the `purchase` dependency and it's not correct to force the user to install it.


adrienpeiffer (approved):

acsonefho:
@OCA/accounting-maintainers Can I have some review here please?

PR #523, opened 2022-07-19
[15.0][MIG] account_payment_term_partner_holiday: Migration to version 15.0

LudLaf:
TT37271

@Tecnativa

Standard migration to version 15.0


LudLaf:
@chienandalu please review.

@pedrobaeza please review.

account_payment_term_partner_holiday/models/account_move.py
```diff
@@ -0,0 +1,44 @@
...
+                self.invoice_date_due = new_invoice_date_due
+
+    def action_post(self):
+        """Inject a context for getting the partner when computing payment term.
+        The trade-off is that we should split the call to super record per record,
+        but it shouldn't impact in performance.
+        """
+        for item in self:
+            _item = item.with_context(move_partner_id=item.partner_id.id)
+            return super(AccountMove, _item).action_post()
```
chienandalu: Not a migration bug but it should be fixed though. This code will allways exit in the first loop iteration. So it will break the expected behavior when several invoices are posted with this method. Attending to the core method design [(odoo/odoo/addons/account/models/account_move.py#L3081-L3086),](https://github.com/odoo/odoo/blob/c79bda62edd8316c2a5754d8aa8838f8aa4caee0/addons/account/models/account_move.py#L3081-L3086) this would be IMO a better approach:
```suggestion
    def action_post(self):
        """Inject a context for getting the partner when computing payment term.
        """
        for move in self:
            super(AccountMove, self.with_context(move_partner_id=move.partner_id.id)).action_post()
        return False
```

ping @victoralmau

chienandalu (changes requested):
Thanks @LudLaf This should be changed thou:

account_payment_term_partner_holiday/models/account_move.py
victoralmau: Thanks, you are right, that is incorrect.

Reviewing in v13 I see that `action_post()` method does not recompute the payment terms lines (this is only done by form when changing some fields), therefore I think that method could be removed directly. What do you think?

chienandalu: What if the invoice date was empty?: https://github.com/odoo/odoo/blob/c79bda62edd8316c2a5754d8aa8838f8aa4caee0/addons/account/models/account_move.py#L2989-L2992

chienandalu (approved):
Let's go with it :)

LudLaf:
All changes have been applied. Thank you for your comments.

victoralmau (approved):

PR #524, opened 2022-07-21
[13.0][FIX] account_payment_term_partner_holiday: Allow action_post() from moves with multi-record

victoralmau:
Allow `action_post()` from moves with multi-record

Related to https://github.com/OCA/account-payment/pull/523#discussion_r925182193

Please @chienandalu and @LudLaf can you review it?

@Tecnativa TT37271

### DIFF
--- a/account_payment_term_partner_holiday/models/account_move.py
+++ b/account_payment_term_partner_holiday/models/account_move.py
@@ -41,4 +41,5 @@ class AccountMove(models.Model):
         """
         for item in self:
             _item = item.with_context(move_partner_id=item.partner_id.id)
-            return super(AccountMove, _item).action_post()
+            super(AccountMove, _item).action_post()
+        return True


account_payment_term_partner_holiday/models/account_move.py
```diff
@@ -41,4 +41,5 @@ def action_post(self):
         """
         for item in self:
             _item = item.with_context(move_partner_id=item.partner_id.id)
-            return super(AccountMove, _item).action_post()
+            res = super(AccountMove, _item).action_post()
+        return res
```
chienandalu: This is confusing as we're overwriting the result value in every loop. I know the original method always returns `True`. Maybe that's what we should return no matter what.

chienandalu (approved):
:smile: :+1:

LudLaf:
Looks good to me

pedrobaeza (approved):

PR #544, opened 2022-09-21
[15.0] [Mig] account_payment_credit_card

Freni-OSI:


bosd (changes requested):
this should be 15.0.1.0.0

PR #545, opened 2022-09-27
[ADD] account_payment_notification: tell people about sent payments

yajo:
This module lets you notify partners about sent payment or debit orders.

You can do that by e-mail and/or by SMS.

@moduon MT-1232


account_payment_notification/views/account_payment_views.xml
```diff
@@ -0,0 +1,100 @@
...
+        />
+        <field name="binding_model_id" ref="account.model_account_payment" />
+        <field name="binding_view_types">form</field>
+        <field name="res_model">mail.compose.message</field>
+        <field name="target">new</field>
+        <field name="view_mode">form</field>
+    </record>
+
+    <record id="action_notifiy_mail_multi" model="ir.actions.act_window">
+        <field name="name">Notifiy by email</field>
```
Shide: Maybe "Notify by email and SMS"?

account_payment_notification/models/account_payment.py
```diff
@@ -0,0 +1,16 @@
...
+class AccountPayment(models.Model):
+    _inherit = "account.payment"
+
+    notification_mobile = fields.Char(related="partner_id.mobile", readonly=True)
+
+    def mark_as_sent(self):
+        """Auto-notify when marking as sent."""
+        result = super().mark_as_sent()
+
+        return result
```
Shide: Is this method empty or I miss something?

Shide (changes requested):

account_payment_notification/views/account_payment_views.xml
yajo: Can't do that, because the wizard is different. However, the user can use the provided filters to choose for example payments for customers that have no email but have phone and notify only those by SMS.

account_payment_notification/models/account_payment.py
yajo: Sorry, it's a remainder I left from the next iteration. I think I'll keep this draft until then, as it's going fast.

yajo:
All comments solved. The module is feature-complete now, ready to review.

yajo:
The coverage problem is due to https://github.com/odoo/odoo/issues/40188. If tested locally after installing `l10n_generic_coa`, tests actually get executed and green.

Not sure how to fix that here though... Any ideas @moylop260 @sbidoul?

rafaelbn:
Reported bugs for fixing in private

account_payment_notification/i18n/es.po
```diff
@@ -0,0 +1,262 @@
...
+"'inbound' else 'payment' }} order of your {{ 'invoice(s)' if object."
+"reconciled_invoices_type == 'invoice' else 'credit note(s)' }} ({{ \", \"."
+"join((object.reconciled_invoice_ids | object.reconciled_bill_ids)."
+"mapped(lambda rec: rec.ref or rec.name)) }}) for the total amount of "
+"{{ format_amount(object.amount, object.currency_id, object.partner_id."
+"lang) }} was processed and sent to the bank. Thanks for your cooperation! -- "
+"{{ object.company_id.name }}"
+msgstr ""
+"Estimado/a {{ object.partner_id.name }}, la orden de {{ 'pago' if object."
+"payment_type == 'inbound' else 'débito' }} de su/s {{ 'factura/s' if object."
```
rafaelbn: @yajo 

Now:
```
"Estimado/a {{ object.partner_id.name }}, la orden de {{ 'pago' if object."
"payment_type == 'inbound' else 'débito' }} de su/s {{ 'factura/s' if object."
```

Should be:
```
"Estimado/a {{ object.partner_id.name }}, la orden de {{ 'cobro' if object."
"payment_type == 'inbound' else 'pago' }} de su/s {{ 'factura/s' if object."
```

account_payment_notification/i18n/es.po
```diff
@@ -0,0 +1,262 @@
...
+"not be monitored nor answered. If you have questions, please contact us by "
+"the usual means.</small>\n"
+"                </p>\n"
+"            </div>\n"
+"        "
+msgstr ""
+"<p>Estimado/a <t t-out=\"object.partner_id.name or ''\">Azure Interior</t>,</"
+"p>\n"
+"            <p t-if=\"object.payment_type == 'inbound'\">\n"
+"                Su orden de débito ha sido procesada y enviada al banco.\n"
```
rafaelbn: @yajo in general in translation "débito" is "cobro" so we should change translation of "debit" with "cobro"

Please review all term and in spanish change "débito" to "cobro"

Thanks! 😄 

Great work!

rafaelbn (commented):
@yajo Some translate issue, functionality perfect! Thank you!

yajo:
Thanks for your in-depth reviews. This is ready to merge now.

rafaelbn (approved):
👍 Reviewed! Thanks!

@Shide , please review 😄

account_payment_notification/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "development_status": "Beta",
+    "category": "Accounting",
+    "website": "https://github.com/OCA/account-payment",
+    "author": "Moduon, Odoo Community Association (OCA)",
+    "maintainers": ["yajo"],
+    "license": "LGPL-3",
+    "installable": True,
+    "depends": [
+        "account",
+        "sms",
```
pedrobaeza, line 16: This dependency and its code should be extracted in a glue module for not making it dependent of such ugly modules and all the IAP infrastructure.

yajo: The fact SMS is installed doesn't mean you have to use it or pay for it. Actually the sms module is open source and allows partners to configure a different provider than Odoo IAP. This module is shipped with SMS notifications disabled by default. But it's part of its core logic and we have no plans or needs to split it.

pedrobaeza: I know, but it's a maintenance burden and doesn't follow usual Odoo modularity, but you can do what you consider.

yajo: Thanks for the suggestion, I understand it. However, for now we'll keep it as it is. The split can be done later by another member that finds that as a requirement, or by ourselves if it becomes a real burden.

account_payment_notification/data/mail_template.xml
```diff
@@ -0,0 +1,49 @@
...
+            </p>
+            <p t-else="">
+                It includes the following credit notes:
+            </p>
+            <ul>
+                <t
+                    t-foreach="object.reconciled_invoice_ids | object.reconciled_bill_ids"
+                    t-as="invoice"
+                >
+                    <li t-out="invoice.ref or invoice.name">INV/2022/00001</li>
```
rafaelbn: Hello @yajo , we have here an issue that we should fix:

- Customer invoices that we debit the number of the invoice is field `name` in model `account.move`, it is not `ref`, never. And field name is mandatory so no needed an "OR"

- Supplier invoices that we credit the number of the invoice is field `ref` in model `account.move`, and name shouldn't be printed because supplier don't know what is that number. Here field ref is not mandatory, so you can say `ref` OR `payment_reference`

So, we need another **if** `inbound` then `name` , **else** `ref` OR `payment_reference`

Thanks

account_payment_notification/data/sms_template.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2022 Moduon Team S.L.
+     License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl). -->
+<data noupdate="1">
+
+    <record id="sms_template_notification" model="sms.template">
+        <field name="name">Payment notification</field>
+        <field
+            name="body"
+        >Dear {{ object.partner_id.name }}, the {{ 'debit' if object.payment_type == 'inbound' else 'payment' }} order of your {{ 'invoice(s)' if object.reconciled_invoices_type == 'invoice' else 'credit note(s)' }} ({{ ", ".join((object.reconciled_invoice_ids | object.reconciled_bill_ids).mapped(lambda rec: rec.ref or rec.name)) }}) for the total amount of {{ format_amount(object.amount, object.currency_id, object.partner_id.lang) }} was processed and sent to the bank. Thanks for your cooperation! -- {{ object.company_id.name }}</field>
```
rafaelbn: @yajo the same issue here `"rec: rec.ref or rec.name"` as above 😢 

- Inbound payments for customer invoices the field is only `name`
- Outbound payment for suppliers invoices the fields are `ref` OR `payment_reference`

If you have any doubt please tell me. I suffer this issue in production.

rafaelbn (commented):
Please review the new issue. Thanks

yajo:
Thanks, all fixed.

rafaelbn (approved):
Perfect! Working in production 👍

rafaelbn:
@yajo we are reviewing with more some users and the last option is not clear

- Notify manually: with this option you have to send emails one by one OR you can also select **massively** all the payments you have to notify and send them by "Action -> Notify partners"
- Notify automatically: the notifications are sent automatically
- Notify automatically, require a notification method: _we don't have clear this option_

imagen

loida-vm:
Reviewed! :+1: 

To improve: add in Readme that emails will be in a waiting queue, and how long it will take to launch.

yajo:
> Notify automatically, require a notification method: _we don't have clear this option_

It means that notifications will be sent automatically and if there's no email or mobile, you'll get an error and won't be able to mark as sent.

loida-vm:
> It means that notifications will be sent automatically and if there's no email or mobile, you'll get an error and won't be able to mark as sent.

So I think that at user level these options should be more understandable.
Add to help popup those descriptions.

yajo:
All attended.

Now there's a separate checkbox for requiring a valid notification method. It has an appropriate help text.



> and how long it will take to launch

I didn't document this because it depends on how the database is configured. Those queue crons are widely known and, if there's any lack of docs, they should be added in their corresponding modules IMHO. However, I did document that the messages are put in the queues.

Shide (approved):
LGTM

loida-vm:
LGTM!

rafaelbn:
@fclementic2c @StefanRijnhart could you please take a look in this PR in order to merge?

I make it easy I've recorder a video with funcionality in runboat 😄 

https://www.loom.com/share/bf7475c9e8df44539e9353aa607957eb

PR #549, opened 2022-10-18
[14.0][FIX] account_payment_return: always using the same return line for reconciling all move lines

eantones:
When the payment return has more than one line, the last payment return line is the only one used to reconcile all the moves on the current payment return. Each return lines should be reconciled only with its related moves.

This bug was introduced in the migration to 14.0 on commit https://github.com/OCA/account-payment/commit/56a98cf49caf0f5df94e52dc1ef6f81a5c157d3f

@joao-p-marques can you please review? Thanks in advance :-)

### DIFF
--- a/account_payment_return/models/payment_return.py
+++ b/account_payment_return/models/payment_return.py
@@ -4,6 +4,7 @@
 # Copyright 2014 Markus Schneider <markus.schneider@initos.com>
 # Copyright 2016 Tecnativa - Carlos Dauden
 # Copyright 2017 Tecnativa - Luis M. Ontalba
+# Copyright 2022 NuoBiT Solutions, S.L. - Eric Antones <eantones@nuobit.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import _, api, fields, models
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
Joao is no longer working with Odoo.

Although not the patch itself, looking at the code, shouldn't the reconciliation be done for the whole set of lines `(move_line2 + return_line.move_line_ids).reconcile()`

eantones:
Probably, but it'd imply to refactor the entire loop and probably the whole method as well. This is just a patch, reconciling the whole set of lines makes sense but I think such refactoring it'd be better to do it in a migration to 15 and 16.

pedrobaeza:
Not really. It's just this:

```python
            for move_line in return_line.move_line_ids:
                # move_line: credit on customer account (from payment move)
                # returned_moves: debit on customer account (from invoice move)
                returned_moves = move_line.matched_debit_ids.mapped("debit_move_id")
                all_move_lines |= move_line
                invoices |= returned_moves.mapped("move_id")
                move_line.remove_move_reconcile()
                return_line.move_line_ids.mapped("matched_debit_ids").write(
                    {"origin_returned_move_ids": [(6, 0, returned_moves.ids)]}
                )
            (return_line.move_line_ids | move_line2).with_context(
                check_move_validity=False
            ).reconcile()
```

eantones:
It would work, let me try it

eantones:
It does not work, it removed all the reconciliations with the "PAY" items as well as the information regarding the refund
image

I added manually the PAY's, revert the changes and it worked again.

Maybe it's less risky if we try to introduce this kind of changes on a migration commit and not on a hotfix.

pedrobaeza (approved):
OK then

omar7r (approved):
Good catch, important fix

PR #555, opened 2022-11-08
[14.0][REF] account_payment_return_import_iso20022: Adapt module to native payment refactoring

pedrobaeza:
Following https://github.com/OCA/bank-payment/pull/979

Blocked by:

- [x] https://github.com/OCA/bank-payment/pull/979

@Tecnativa

### DIFF
--- a/account_payment_return_import_iso20022/models/payment_return.py
+++ b/account_payment_return_import_iso20022/models/payment_return.py
@@ -1,5 +1,6 @@
 # Copyright 2016 Tecnativa - Carlos Dauden
 # Copyright 2020 Tecnativa - Víctor Martínez
+# Copyright 2022 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import models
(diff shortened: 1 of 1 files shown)


account_payment_return_import_iso20022/models/payment_return.py
```diff
@@ -9,10 +10,20 @@ class PaymentReturnLine(models.Model):
...
-        lines = self.filtered(lambda x: not x.move_line_ids and x.reference)
-        for line in lines:
-            line.move_line_ids = AccountMoveLine.search(
-                [("bank_payment_line_id.name", "=", line.reference)]
+        """Include in the matches the lines coming from payment orders."""
+        matched = self.env["payment.return.line"]
+        for line in self.filtered(lambda x: not x.move_line_ids and x.reference):
+            payments = self.env["account.payment"].search(
+                [
+                    ("payment_reference", "=", line.reference),
```
carlosdauden: payment_reference is not sended as end2end_identification:
https://github.com/OCA/bank-payment/pull/979/files#diff-514ea70d8407133fa56949b0111cc1ff8befe4aea3d5160c6863f2099a8927aaR159

EndToEndId is the one used by the iso20022 scheme to indicate which payment is returned and for this reason it is  established as a reference when importing the file:
https://github.com/OCA/account-payment/blob/c1f620c97127d2fa6191f6464165685355a5e17c/account_payment_return_import_iso20022/wizard/pain_parser.py#L62-L64

The implemented search may be fine as an alternative search, but the main search must be based on what is sent with the EndToEndId tag.

carlosdauden (changes requested):

account_payment_return_import_iso20022/models/payment_return.py
pedrobaeza: OK, I didn't know about the usage of `EndToEndId`. Then it's going to be even easier. I'll change it.

pedrobaeza: Fixed, and added compatibility with old approach (both in `account_payment_order` and here).

fredaspen1:
> Following https://github.com/OCA/bank-payment/pull/979
> 
> Blocked by:
> 
> - [x] https://github.com/OCA/bank-payment/pull/979
> 
> @Tecnativa 

Hi Pedro, we are looking for dev, to build a payment platform for banks using crypto as a way of payment ans comply with the iso20022 standard

PR #577, opened 2023-01-23
[13.0][ADD] account_payment_term_security: New module

victoralmau:
New module

Please @CarlosRoca13 and @pedrobaeza can you review it?

@Tecnativa TT41273


CarlosRoca13 (approved):

account_payment_term_security/security/security.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="account_payment_term_mgmt" model="res.groups">
+        <field name="name">Enable to manage payment terms</field>
+        <field name="category_id" ref="base.module_category_accounting_accounting" />
```
pedrobaeza: This can't be in accounting, as it creates that this is no more a dropdown. It should be on the technical category.

account_payment_term_security/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This addon creates a new group called "Enable to manage payment terms" in Invoicing.
```
pedrobaeza: Please extract part of this in CONFIGURE, and do the description more "commercial".

account_payment_term_security/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
...
+
+This new group allows to manage payment terms in partners.
```
pedrobaeza: Talk here about restricting the change of the payment terms in the partners.

victoralmau:
Changes done.

pedrobaeza (changes requested):
The payment term should be visible for the rest of the users, but not editable. And you should do the same on invoices.

victoralmau:
> The payment term should be visible for the rest of the users, but not editable. And you should do the same on invoices.

Ok, changes done (and also applied to invoices).

account_payment_term_security/README.rst
```diff
@@ -0,0 +1,83 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/account-payment-13-0/account-payment-13-0-account_payment_term_security
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/96/13.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+In partners and invoices, payment terms can only be managed when the user has permission "Enable to manage payment terms".
```
pedrobaeza:
```suggestion
In partners and invoices, payment terms can only be changed when the user has permission "Change payment terms in partners and invoices".
```

account_payment_term_security/README.rst
```diff
@@ -0,0 +1,83 @@
...
+
+.. contents::
+   :local:
+
+Configuration
+=============
+
+To use this module, you need to :
+
+Go to Settings > Users & Companies > Users (edit your user) > Set "Enable to manage payment terms" option.
```
pedrobaeza:
```suggestion
Go to Settings > Users & Companies > Users (edit your user) > Set "Change payment terms in partners and invoices" option.
```

account_payment_term_security/models/account_move.py
```diff
@@ -0,0 +1,31 @@
...
+from lxml import etree
+
+from odoo import api, models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    @api.model
+    def fields_view_get(
```
pedrobaeza, line 14: Why do it dynamically instead of putting 2 times the same field, one with readonly and the other without, with the corresponding groups?

account_payment_term_security/models/res_partner.py
```diff
@@ -0,0 +1,38 @@
...
+from lxml import etree
+
+from odoo import api, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    @api.model
+    def fields_view_get(
```
pedrobaeza, line 14: Same question

account_payment_term_security/models/account_move.py
victoralmau: I have preferred to make this solution to avoid:
- Having to define the new field with what I had previously defined and other modules could add (groups, readonly, attrs, etc).
- If another module wants to use the field selector to change something it will be a problem because it will have to do it twice (the existing field and the new one).

Anyway, if you think it's better to add the field in the view and have it appear twice, no problem.

pedrobaeza: OK, I see. Let's keep it this way.

pedrobaeza (approved):

PR #582, opened 2023-02-01
[14.0][FIX] account_payment_return_import_iso20022: Get the correct line to return

### LINKED ISSUE #581: [14.0] account_payment_return_import_iso20022: Not returns the correct move line
Hi

From last commit https://github.com/OCA/account-payment/commit/1667fb75e79fdb152cce4a8558fb06dcb0dd40c3 caused by the refactoring of payment orders, when finding matches for returns, it gets the line of the pending account of journal instead of customer's account, that is the account that we need to return for opening the invoice again.

I think should be replaced
`                    line.move_line_ids |= payment.move_id.line_ids.filtered(
                        lambda x: x.account_id != payment.destination_account_id
                        and x.partner_id == payment.partner_id
                    )`

by:

 `                    line.move_line_ids |= payment.move_id.line_ids.filtered(
                        lambda x: x.account_id == payment.destination_account_id
                        and x.partner_id == payment.partner_id
                    )`

@pedrobaeza What do you think?
Regards

omar7r:
In addition it sets the partner_id in the payment order line again

### DIFF
--- a/account_payment_return_import_iso20022/models/payment_return.py
+++ b/account_payment_return_import_iso20022/models/payment_return.py
@@ -24,10 +24,11 @@ class PaymentReturnLine(models.Model):
                 ],
             )
             if payments:
+                line.partner_id = payments[0].partner_id
                 matched += line
                 for payment in payments:
                     line.move_line_ids |= payment.move_id.line_ids.filtered(
-                        lambda x: x.account_id != payment.destination_account_id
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Not tested, but trusting on the contributor

carlosdauden (approved):

PR #585, opened 2023-02-06
[14.0][IMP] account_payment_return: Return to same account used in payment receipt

carlosdauden:
If payment account used is payment_debit_account_id this account should be used in the payment return

@Tecnativa

ping @omar7r @ao-landoo


pedrobaeza (approved):

omar7r (approved):
LGTM

PR #588, opened 2023-02-16
[MIG] account_cash_discount_payment: Migration to 16.0

AnizR:
This migration is done considering that we use odoo/account instead of oca/account_cash_discount_base to manage discounts.


dsolanki-initos (changes requested):
Could you please arrange the commit as per the OCA guidelines?

While functional testing got an error. See the below image.

image

AnizR:
> Could you please arrange the commit as per the OCA guidelines?
> 
> While functional testing got an error. See the below image.
> 
> image

I tried to reproduce your "bug" but I don't understand how you managed to have an account.move.line without an account.move linked to it (field move_id is required).

PR #589, opened 2023-02-16
[14.0][FIX] account_payment_batch_process (Discount Taken Issue)

patrickrwilson:
The discount taken on the invoice wasn't being set and here is a fix.

### DIFF
--- a/account_payment_batch_process/wizard/account_payment_register.py
+++ b/account_payment_batch_process/wizard/account_payment_register.py
@@ -422,6 +422,28 @@ class AccountPaymentRegister(models.TransientModel):
                 ).reconcile()
             if any(
                 group_data[partner]["inv_val"][inv.id]["payment_difference_handling"]
+                == "reconcile"
+                for inv in invoices
+            ):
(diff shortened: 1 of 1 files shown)


account_payment_batch_process/wizard/account_payment_register.py
```diff
@@ -420,6 +420,28 @@ def make_payments(self):
...
+                == "reconcile"
+                for inv in invoices
+            ):
+                for inv in invoices:
+                    if (
+                        group_data[partner]["inv_val"][inv.id][
+                            "payment_difference_handling"
+                        ]
+                        == "reconcile"
+                    ):
```
dreispt: Here is how you can avoid the duplicate logic and still not have unnecessary loops.
```suggestion
            invoices_diff_reconcile = invoices.filtered(
                lambda inv:
                group_data[partner]["inv_val"][inv.id]["payment_difference_handling"]
                == "reconcile"
            )
            for inv in invoices_diff_reconcile:
```

PR #600, opened 2023-03-03
[14.0][IMP] pre-commit update

Aldeigja:
Updated pre-commit to the latest version


pedrobaeza:
Merging as the error in tests is not due to this (although it should be checked).

PR #611, opened 2023-03-22
[16.0][MIG] account_payment_return: Migration to 16.0

ao-landoo:
Standard migration

@carlosdauden


account_payment_return/tests/test_payment_return.py
```diff
@@ -42,11 +39,9 @@ def setUpClass(cls):
             {
                 "name": "Test income account",
                 "code": "INCOME",
-                "user_type_id": cls.env["account.account.type"]
-                .create(
+                "account_type": "asset_receivable".create(
```
carlosdauden: Remove this create

carlosdauden (changes requested):

account_payment_return/tests/test_payment_return.py
ao-landoo: Done. There is one last test failing. It's weird because invoice_payments_widget field's value is False once the payment return is confirmed.

Edit: Looks right. Once payment is reverted Odoo's not displaying payment information on widget

account_payment_return/tests/test_payment_return.py
```diff
@@ -232,20 +225,13 @@ def test_duplicate_lines(self):
             )
 
     def test_payments_widget(self):
-        info = json.loads(self.invoice.invoice_payments_widget)
+        info = self.invoice.invoice_payments_widget
         self.assertEqual(len(info["content"]), 1)
         self.payment_return.action_confirm()
-        info = json.loads(self.invoice.invoice_payments_widget)
-        self.assertEqual(len(info["content"]), 2)
-        self.assertEqual(info["content"][1]["amount"], -500.0)
```
carlosdauden: Why do you remove tests?
Have you tested the behavior in previous versions? Does it behave the same?

carlosdauden (changes requested):

account_payment_return/tests/test_payment_return.py
ao-landoo: Widget information has been changed from V15 to V16, now if payment is reverted widget shows no information, so those test lines have no sense now.

pedrobaeza: Then the alternative code for the new widget should be developed and tested here, but you can't rip-off features.

ao-landoo: I could add something like

`self.assertEqual(len(info["content"]), 0)`

That's the only thing the widget does if the payment has been revered. If you don't mean that I'm not getting your point

pedrobaeza: I'm afraid not. What I mean is that the module should add the return information on the invoice footer where the payments appear. Is such feature restored in this migration with whatever code needed? If so, then test here that part. If not, you should re-develop such feature and test it here.

Yadier-Tecnativa:
supersede by https://github.com/OCA/account-payment/pull/639

PR #624, opened 2023-04-18
[15.0][FIX] account_payment_promissory_note: avoid error when paying vendor bills

oihane:


pedrobaeza (approved):
Fast-tracking this, as it's an obvious patch for avoiding the error `max(False)`:

/ocabot merge patch

PR #625, opened 2023-05-05
[15.0][FIX] account_payment_return_import_iso20022: Restore compatibility with pre-payment orders

pedrobaeza:
This legacy compatibility shouldn't be removed yet, as if you migrate from <v14, you don't have the new references.

@Tecnativa TT42966


pedrobaeza:
Fast-tracking, as this is something incorrectly rip-off:

/ocabot merge patch

PR #635, opened 2023-05-29
[15.0][FIX] account_payment_return_import_iso20022: Method _compute_amount is not called…

carlosdauden:
… because records are removed in super()

In the original method all the records are expected:
https://github.com/OCA/account-payment/blob/ad9ee855fed524c0d7bf216d9a289aa2fb5ec272/account_payment_return/models/payment_return.py#L377-L392

@Tecnativa TT43373


pedrobaeza (approved):
Can you please add to the commit message and PR the steps to reproduce the problem?

sergio-teruel (approved):

PR #646, opened 2023-06-23
[16.0][MIG] account_due_list_days_overdue

ChrisOForgeFlow:
@ForgeFlow


mileo:
Could you reopen?

MiquelRForgeFlow (approved):

account_due_list_days_overdue/models/account_move_line.py
```diff
@@ -0,0 +1,119 @@
...
+    @api.model
+    def fields_view_get(
+        self, view_id=None, view_type="form", toolbar=False, submenu=False
+    ):
+        result = super().fields_view_get(
+            view_id,
+            view_type,
+            toolbar=toolbar,
+            submenu=submenu,
+        )
```
rvalyi: this fields_view_get method does not exist anymore in Odoo v16 models.py so I'm afraid this override won't be called and it needs to be migrated, Usually this is replaced by the new `def _get_view(self, view_id=None, view_type="form", **options):` in v16...

rvalyi (changes requested):

account_due_list_days_overdue/models/account_move_line.py
ChrisOForgeFlow: I changed to new function, I suppose use get_view instead _get_view function is not an issue in implementation

JordiBForgeFlow:
@rvalyi I can see that you requested changes and they have been attended. Can you update your review? We're hoping to get this one merged.

matteoopenf:
@rvalyi  any news?

rvalyi (approved):
sorry for the delay, I think the migration is OK now with the changes in view_get

matteoopenf:
> sorry for the delay, I think the migration is OK now with the changes in view_get

Thank you

PR #647, opened 2023-06-28
[15.0][MIG] account_check_printing_report_dlt103

JasminSForgeFlow:
Standard Migration

@ForgeFlow


AaronHForgeFlow (approved):
Functional tests

account_check_printing_report_dlt103/__manifest__.py
```diff
@@ -16,7 +15,6 @@
     "data": [
         "data/report_paperformat.xml",
         "data/report_paperformat_parameter.xml",
-        "data/account_payment_check_report_data.xml",
```
MiquelRForgeFlow: Why did you delete this?

MiquelRForgeFlow (approved):
LGTM

account_check_printing_report_dlt103/__manifest__.py
JasminSForgeFlow: That file is used to add option dlt103 for check printing, which is handled with selection field account_check_printing_layout on payment.journal in v15, so no longer a need for that file with that declaration way

PR #639, opened 2023-06-14
[16.0][MIG] account_payment_return: Migration to version 16.0

Yadier-Tecnativa:
@Tecnativa TT43212

PR #649, opened 2023-07-04
[16.0][FIX] account_payment_return: Refactor wrong code

Raul-S73:
FIX sigleton error when print 2 or more invoices:
image

Added in #639

### DIFF
--- a/account_payment_return/models/account_move.py
+++ b/account_payment_return/models/account_move.py
@@ -53,10 +53,11 @@ class AccountMove(models.Model):
         }
 
     def _compute_payments_widget_reconciled_info(self):
-        if not self.returned_payment:
-            return super()._compute_payments_widget_reconciled_info()
-        else:
-            for move in self:
+        moves_to_compute = self.env["account.move"]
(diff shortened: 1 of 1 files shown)


account_payment_return/models/account_move.py
```diff
@@ -53,10 +53,11 @@ def prepare_values_returned_widget(self, line_id, amount, is_return=False):
         }
 
     def _compute_payments_widget_reconciled_info(self):
-        if not self.returned_payment:
-            return super()._compute_payments_widget_reconciled_info()
-        else:
-            for move in self:
+        moves_to_compute = self.env["account.move"]
+        for move in self:
+            if move.returned_payment:
```
pedrobaeza:
```suggestion
            if not move.returned_payment:
```

Raul-S73: @pedrobaeza Done!

pedrobaeza (approved):

Yadier-Tecnativa (approved):

PR #658, opened 2023-07-28
[15.0] [account_payment_widget_amount] - Fix account payment widget amount

hieulucky111:
When Invoice has Payment Term with Multiple Payment period. Partial amount will be wrong in some cases

### DIFF
--- a/account_payment_widget_amount/models/account_move_line.py
+++ b/account_payment_widget_amount/models/account_move_line.py
@@ -36,7 +36,11 @@ class AccountMoveLine(models.Model):
             decimal_places = current_am.company_id.currency_id.decimal_places
             new_partials = []
             for partial in partials:
-                if current_am.currency_id.id != current_am.company_currency_id.id:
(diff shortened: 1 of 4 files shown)


account_payment_widget_amount/models/account_move_line.py
```diff
@@ -1,5 +1,3 @@
-# Copyright 2018-2021 ForgeFlow S.L.
```
ChrisOForgeFlow: You should not remove copyright section

ChrisOForgeFlow (commented):
Can you add please test case, to understand cases when algorithm goes wrong?

account_payment_widget_amount/models/account_move_line.py
hieulucky111: Ok, i am working on some new test cases.

ChrisOForgeFlow (approved):
LGTM+

PR #672, opened 2023-10-04
[ADD]account_due_list_edit_inline: editing name and date inline

odooNextev:


GSLabIt (approved):

francesco-ooops (approved):
Functional ok!

francesco-ooops:
@OCA/banking-maintainers could this be merged? thanks!

francesco-ooops:
@etobella anything against this?

etobella:
Why aren't you adding it directly to the base module? It hasn't have too much sense IMO.

francesco-ooops:
@etobella we didn't want to change formatting of "date" column single-handedly for everyone... but we wouldn't mind adding it to base module. 

Do you think we should add an extra column and let user choose which one to use? or what do you suggest?

francesco-ooops:
@etobella shall we merge as it is?

francesco-ooops:
@etobella ping :)

etobella:
Yes, I think we should merge it as a lot of modules have been done for this... I would propose to merge them in a single module on 16 or 17, but not right now....

/ocabot merge nobump

PR #679, opened 2023-10-18
[10.0][MIG] account_payment_notification backport from v15

AaronHForgeFlow:
Removed sms feature as long sms is not available in v10

cc @ForgeFlow


AaronHForgeFlow:
simplify

PR #703, opened 2024-01-26
[12.0][UPD] Repo template

alexey-pelykh:


alexey-pelykh:
@pedrobaeza 😩 what should I do with modules maturity that are outside the repo?

pedrobaeza:
Disable the maturity check here for this version. There's a question about that on the copier template.

PR #718, opened 2024-03-13
[17.0] [MIG] account_due_list migration to 17.0

luisDIXMIT:
Standard migration

@etobella


etobella (approved):
Tested locally

:+1:

account_due_list/views/payment_view.xml
```diff
@@ -41,6 +40,7 @@
                 />
                 <field name="debit" readonly="1" sum="Total Debit" optional="hide" />
                 <field name="credit" readonly="1" sum="Total Credit" optional="hide" />
+                <field name="company_currency_id" invisible="1" readonly="1" />
```
pedrobaeza, line 44: Why do you need to add this?

luisDIXMIT: Because without the company currency the sum totals fields cannot be displayed.

Without the company_currency_id:
image


With the company_currency_id:
image

pedrobaeza: Thanks for the explanation. Then put a comment above the line indicating that. If not, someone (or your future you) will be tempted to remove the line thinking it's useless.

pedrobaeza (approved):
/ocabot migration account_due_list
/ocabot merge nobump

PR #720, opened 2024-03-18
[16.0][FIX] account_payment_line, wrong on outbound and writeoff

kittiu:
Before this fix, things (sign + or -) seem to go wrong on Supplier Payment side, especially when use together with write off.

For example,

1. When Propose Payment Distribution, the allocation amount is (+), while I think this should be (-) as other columns (otherwise it is very confusing)

image

2. Then I click on Full Paid the writeoff amount become -200

image

3. Or I choose to allocate only partial 90 out of 100, and tick Full Paid again, I would expect to write off = -10, but it is -190

image

4. Or I redo again, 1) Amount 190, 2) system allocate, tick full paid 3) click Confirm

image

I also wonder why there are so much bugs here. But I think the previous author focus on testing for the Customer Payment side, which seem to be OK.

So, this PR is trying to make it works on both side.


kittiu:
@ChrisOForgeFlow can you help check if I am not misunderstanding your design intention.

kittiu:
move to #721

PR #723, opened 2024-04-09
[15.0][IMP] account_check_report many fixes and improvements

AaronHForgeFlow:
[FIX] account_check_report: payment amount should consider current payment and not other payments

[IMP] account_check_report: show amount residual at the payment date, not on today's date

[IMP] account_check_report: show direct refund applied on invoices paid by the check e.g. 2 invoices and a refund for those 2 invoices should appear, even if the refund includes separate invoices

[IMP]account_check_report: button invoices show all related invoices + direct refunds

[IMP] account_check_report: add test coverage

[IMP] account_check_report: do not print more than one line by invoice. Currently, if there are several receivable/payable lines, for example, due to payment terms the invoice appears multiple times

[FIX] Do not show "Check: False" when there is no check number, do not show anything instead.

Kind of forward porting this #397  --> Screenshots for the different fixes are provided there

The changes in the payment document itself can be reverted if only the report is accepted.

### DIFF
--- a/account_check_report/__init__.py
+++ b/account_check_report/__init__.py
@@ -1,1 +1,2 @@
 from . import report
+from . import models
--- a/account_check_report/models/__init__.py
+++ b/account_check_report/models/__init__.py
@@ -1,0 +1,1 @@
+from . import account_payment
--- a/account_check_report/models/account_payment.py
+++ b/account_check_report/models/account_payment.py
@@ -1,0 +1,92 @@
+# 2016-2024 ForgeFlow S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import api, models
+from odoo.osv import expression
+
+
+class AccountPayment(models.Model):
+    _inherit = "account.payment"
+
+    @api.depends(
+        "move_id.line_ids.matched_debit_ids", "move_id.line_ids.matched_credit_ids"
+    )
+    def _compute_stat_buttons_from_reconciliation(self):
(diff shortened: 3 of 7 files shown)


AaronHForgeFlow:
One of the main needs for this fix is the first one:

[FIX] account_check_report: payment amount should consider current payment and not other payments

For example, when having to bills, for $2.99 each and 2 payments for $1 and $ 4 where the matching is the following:

bill1: reconcile with payment 1 --> amount residual left $1.99  (payment 1 applied $1)
bill1: reconcile with payment 2--> amount residual left $0  (payment 2 applied $1.99)
bill2: reconcile with payment 2--> amount residual left $0.98 (payment 2 applied $2.01)

When we print the report on the first payment (right part of the following image) we see that the payment amount is $2.99. This is incorrrect, as only 1$ was applied. The left part of the image shows how it looks after this fix

check_report_paymnet1

Same thing with the report for for the second payment. It pays $1.99 on the first invoice, not the total amount:
check_report_paymnet2

AaronHForgeFlow:
🎶🐛🎶 "cri-cri... cri-cri... criiiiiiick!" 🐜🎶🐛  
🐜 "cricket chirp sounds intensify..." 🎶🐛 "cri-criiiiick cri-cri!" 🎶🐜  
🐛🐜🎶 "cri-crick cri-crick!" 🎶🐜🐛

MiquelRForgeFlow:
Who truly makes 🎶 cri-cri 🎶 is the 🪳🦗

PR #724, opened 2024-04-19
[14.0] [FIX] account_payment_term_discount: Difference in rounding

majouda:
In the context of this module's application, we encountered a specific issue involving:

* Discount Amount: 229.95
* Payment Difference: 229.95000000000002

Although there is no rounding in _compute_payment_difference() of account module and this module performs rounding in _compute_discount_amt(). 

This discrepancy in rounding can lead to different comparison results. 

This commit aims to rectify the inconsistency in rounding between these two fields, ensuring accurate comparisons.

### DIFF
--- a/account_payment_term_discount/wizard/account_payment_register.py
+++ b/account_payment_term_discount/wizard/account_payment_register.py
@@ -4,7 +4,6 @@ from dateutil.relativedelta import relativedelta
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
-from odoo.tools import float_round
 
 
 class AccountPaymentRegister(models.TransientModel):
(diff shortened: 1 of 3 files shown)


bodedra:
@majouda Thanks for your contribution. Would you add screenshot or example before and after this PR changes, please?

account_payment_term_discount/wizard/account_payment_register.py
```diff
@@ -45,7 +45,7 @@ def onchange_payment_amount(self):
                 payment_date = fields.Date.from_string(self.payment_date)
                 discount_amt = self.invoice_id.discount_amt
 
-                payment_difference = self.payment_difference
+                payment_difference = abs(round(self.payment_difference, 2))
```
hildickethan: it would be better to use `float_round` with the `invoice_id.currency_id`'s precision, ultimately the same result for most but it would save headaches for any currency that doesn't use 2 decimal places

majouda: Hello @hildickethan-S73 I made some changes according to your comment. 

Can you please check again?

majouda:
> @majouda Thanks for your contribution. Would you add screenshot or example before and after this PR changes, please?


Hello @bodedra I added a description to this PR, can you check plz if it's clear now?

hildickethan (approved):

bodedra (approved):
LGTM :+1:

PR #734, opened 2024-06-05
[17.0][FIX] account_payment_term_extension: Add selection options translatable

ramiadavid:


pedrobaeza (approved):
Please take the occasion to rename the method to the guideline `_selection_delay_type`.

ramiadavid:
@pedrobaeza Why don't all the terms to translate appear on weblate? In the .pot file they do appear, but in the .po of the languages, there are many that do not appear, and I don't know how to add them, since it says that it is 100% translated when it is not.
image
image

pedrobaeza:
Can you mention one specifically? The bot, after a merge, regenerate the POT file, and then reapply the POT on each .po, so they should be synchronized. Maybe something in the middle has failed.

ramiadavid:
In the images you can see that in English there are 46 terms vs 33 in Spanish
For example "Days and Months" in spanish in Spanish it is not

pedrobaeza:
@sbidoul do you know why there's no PO update after POT regeneration?

pedrobaeza (approved):

neo-oien:
Hello,

I've detected the same problem. 
Option values for 'delay_type' field don't appear in .PO file and there is no way to translate them.

imagen

Best regards

PR #740, opened 2024-06-24
[16.0][FIX] account_payment_batch_process: allow original domain of payment method line

JordiBForgeFlow:
As the existing domain is good enough to limit the choices of payment methods, and the previous approach would be confusing because it would show payment method lines associated with all the journals.


JordiBForgeFlow:
@alexpforgeflow @leemannd

AlexPForgeFlow (approved):
LGTM!

LoisRForgeFlow (approved):
Nice to reduce code when possible :+1:

PR #744, opened 2024-07-02
[16.0][MIG] account_payment_term_partner_holiday: Migration to 16.0

PabloCatalinaCastello:
- Reopen  #636


account_payment_term_partner_holiday/models/account_move.py
```diff
@@ -0,0 +1,43 @@
...
+                self.invoice_date_due
+            )
+            if new_invoice_date_due != self.invoice_date_due:
+                self.invoice_date_due = new_invoice_date_due
+
+    def action_post(self):
+        """Inject a context for getting the partner when computing payment term."""
+        for move in self:
+            super(
+                AccountMove, self.with_context(move_partner_id=move.partner_id.id)
```
alvaro-domatix: If self is being iterated the super should be done for "move.with_context(..." not "self.with_context(..."

alvaro-domatix (changes requested):

alvaro-domatix (approved):

alvaro-domatix:
Hi @pedrobaeza , could you please take a look at this PR when you have a chance? Thanks!

pedrobaeza:
Sorry, not using this module.

/ocabot migration account_payment_term_partner_holiday

alvaro-domatix:
Sorry for being curious, but can I ask why? Is it just a coincidence, or is there another reason?

Thanks Pedro!

pedrobaeza:
We don't have customers with this need for now. I also personally think that a payment debt shouldn't be delayed because the debtor is on holidays. It's the payer problem. If you go on holiday, program the payments.

nachoserra-domatix (approved):
lgtm!

Gerloav:
LGTM!

percevaq:
When you have a client that generates invoices for temporary cycles to their customers, they have to take into account those vacation days to generate the due dates when they are sure that they will be attended.

LGTM!

antoniospneto (approved):
I understand this can be a bit inconvenient, but yes, this is cultural here in Brazil. Some clients get upset if the due date falls during their holidays."

dreispt (changes requested):
If this is a migration, the commit history is missing.

PR #746, opened 2024-07-08
[13.0][FIX] account_cash_discount_write_off: use correct tags when using refunds

AnizR:
The computation of tags of taxes can be wrong when doing a payment of an invoice with refunds.

When setting tags, the type of the account_move (refund or not) should directly come from the invoice that will be generated.


rousseldenis (approved):
Code review

PR #747, opened 2024-07-16
[16.0][FIX] account_cash_invoice: Do not create account for test

carlos-lopez-tecnativa:
- This account code conflicts with other chart accounts(other modules), so we are removing it.

https://github.com/odoo/odoo/blob/16.0/addons/l10n_hr_euro/data/account.account.template.csv#L325
https://github.com/odoo/odoo/blob/16.0/addons/l10n_fr/data/account.account.template.csv#L368

Log
`psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "account_account_code_company_uniq"﻿[﻿2077﻿]DETAIL: Key (code, company_id)=(449000, 1) already exists.`

@tecnativa TT50198


pedrobaeza (approved):
Let's avoid conflicts in integration tests.

Fast-tracking this:

/ocabot merge patch

PR #748, opened 2024-07-16
[16.0][FIX] account_cash_invoice: Filter only lines not reconciled

carlos-lopez-tecnativa:
With other modules, it raises the error: "You are trying to reconcile some entries that are already reconciled."

@tecnativa TT50198


pedrobaeza (approved):
/ocabot merge patch

PR #751, opened 2024-07-23
[16.0][MIG] account_payment_credit_card: Migration to version 16.0

carlos-lopez-tecnativa:
@Tecnativa TT49796


carlos-lopez-tecnativa:
@carolinafernandez-tecnativa  @pedrobaeza could you please help me with the review of this module? Thank you!

account_payment_credit_card/views/account_view.xml
```diff
@@ -0,0 +1,24 @@
...
+    <record id="view_account_journal_creditcard_form" model="ir.ui.view">
+        <field name="name">account.journal.form</field>
+        <field name="model">account.journal</field>
+        <field name="type">form</field>
+        <field name="inherit_id" ref="account.view_account_journal_form" />
+        <field name="arch" type="xml">
+            <field name="default_account_id" position="after">
+                <field
+                    name="credit_card"
+                    attrs="{'readonly':[('type', '!=', 'bank')]}"
```
carolinafernandez-tecnativa: Maybe it is not necessary to show this fields if it not journal type bank

image

account_payment_credit_card/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,3 @@
+* Go to Accounting > Configuration > Journals
+* Create a new journal for your credit card
```
carolinafernandez-tecnativa: Please change to: Create a new journal bank for your credit card

carolinafernandez-tecnativa (commented):
Check this example:

image

I registered payment to this invoice with journal configuration

image

Payment is created correctly, but invoice still pending to paid and let me add many payments...

image

If you test it with the same partner as set in journal, it happens the same.
Is this behavoiur is on the goal of this module?

Journal configuration:

image

carlos-lopez-tecnativa:
@carolinafernandez-tecnativa I noticed that you tested with a `customer invoice`, but this module is intended only for `vendor bills`. Your images show inconsistencies; for instance, the first invoice is `INV/2024/00006`, but the payment is linked to invoice `INV/2024/00005`. 

Please test with vendor bills. 

@pedrobaeza, could you provide your comments on whether we need to add a constraint to prevent usage with customer invoices?

With vendor bills, it works fine for me.
image
image

carolinafernandez-tecnativa:
Ok, please check with customer invoices is making an unusual behavour that was my intention to transmit for journal that has the configuration. Maybe add a restriction not to use this journal to pay customer invoices.

carlos-lopez-tecnativa:
> Ok, please check with customer invoices is making an unusual behavour that was my intention to transmit for journal that has the configuration. Maybe add a restriction not to use this journal to pay customer invoices.


Yes, this behavior has been inconsistent since version 14 (it only worked for vendor bills). @pedrobaeza, should we add this restriction as part of the migration, or should we not add the constraint?

pedrobaeza:
/ocabot migration account_payment_credit_card

I don't think we should restrict them, as there can be legit debits in these journals. The README is clear stating about vendor bills, so nothing more to be done on your side.

pedrobaeza (approved):
Code review

victoralmau (approved):
Code review OK

PR #752, opened 2024-07-24
[17.0][MIG] account_payment_return: Migration to version 17.0

carlos-lopez-tecnativa:
@Tecnativa TT49281


carolinafernandez-tecnativa:
I have tested on runboat and only one comment:

If i duplicate one customer payment return confirmed, Reference to the created journal entry is copied from original customer payment return. This field should be empty when copying it.

image

carlos-lopez-tecnativa:
> I have tested on runboat and only one comment:
> 
> If i duplicate one customer payment return confirmed, Reference to the created journal entry is copied from original customer payment return. This field should be empty when copying it.

@carolinafernandez-tecnativa  Good catch! I made the change. Please update your review. Thanks!

carolinafernandez-tecnativa (approved):
LGTM Thanks :)

smcqueen81 (approved):
Ok functional

carolinafernandez-tecnativa:
ping @pedrobaeza

pedrobaeza (approved):
Thanks for the improvements and fixes:

/ocabot merge nobump

You can backport to 15/16 the copy=False patch.

PR #757, opened 2024-08-09
[16.0][FIX] account_payment_return: don't copy some fields

carlos-lopez-tecnativa:
If an invoice with the returned_payment field set is duplicated, the new invoice will incorrectly have this field activated. This fix prevents that from happening. same case for field move_id on payment.return

Partial backport of https://github.com/OCA/account-payment/pull/752 according comment https://github.com/OCA/account-payment/pull/752#pullrequestreview-2199342339

@Tecnativa TT49281

@pedrobaeza @carolinafernandez-tecnativa  could you review please


carolinafernandez-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #758, opened 2024-08-30
[16.0][FIX] account_payment_term_extension: Add cash_rounding parameter to _compute_terms() to prevent error

victoralmau:
Add cash_rounding parameter to `_compute_terms()` to prevent error

Required since https://github.com/odoo/odoo/commit/84b03f7e90f49762612875ac61a8fb29920c585d

Please @pedrobaeza can you review it?

@Tecnativa

### DIFF
--- a/account_payment_term_extension/README.rst
+++ b/account_payment_term_extension/README.rst
@@ -71,6 +71,11 @@ Select any payment term and set a date in invoice.
 
 You must see the due date based on this payment term.
 
+Known issues / Roadmap
+======================
+
+This module is not compatible with cash rounding
+
 Bug Tracker
 ===========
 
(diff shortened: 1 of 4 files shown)


pedrobaeza:
This module seems a bit bloated and copying/pasting a lot of code from upstream, while the feature it brings don't require it directly. Can't this be rethought for not requiring this doing some kind of tricks?

victoralmau:
I agree that currently the `_compute_terms()` method copies a lot of code and probably can be refactored but, is it the time now? I think not, what is necessary now is to unblock those who use this module and then, create a new PR to refactor this if possible (probably it will be necessary to create extra methods in odoo core).

pedrobaeza:
Then just add the parameter, and issue an error if the parameter is used, adding in the ROADMAP that this module is not compatible with cash rounding.

pedrobaeza (approved):

pilarvargas-tecnativa (approved):

PR #775, opened 2024-12-05
[16.0][ADD] payment_partner

etobella:
This module allows to set global and non global payment providers. Then, we can specify the non global providers on each partner.


payment_partner/README.rst
```diff
@@ -0,0 +1,88 @@
...
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+1. Access the Payment Providers page
+2. Unset the is global selector
```
luisDIXMIT, line 42:
```suggestion
2. Set filter mode to Include or Exclude
```
If filter mode is unset or set to global we can't do the next step because partners field will be invisible

luisDIXMIT (approved):
Tested functionally on local and LGTM. Only one minor suggestion.

flotho (approved):
awesome feature, thanks
code reviwe,
LGTM

PR #780, opened 2025-01-14
[16.0][FIX] account_payment_term_extension: Display correct amounts in journal items when using foreign currency

sergiobstoj:
Detected problem:

When creating an invoice for 900EUR (in a company currency with USD) with a Payment Term with multiple terms, for example **30% now, the rest in 60 days**, the journal items that are generated are displaying the amount in EUR but with USD currency.

image

Specifically, total invoice amount in EUR is 900, and converted to USD is 1376.
First payment should be 30% -> 412,8 USD
Second payment should be balance -> 963,2 USD

Currently, this info is displayed as:
image

Payment terms are being computed considering EUR currency.
As:
- First payment, 30% -> 270
- Second payment, balance -> 1106

Correct computing should be in company's currency. This PR fixes that:
image


aritzolea:
Could you add a unit test that would not pass without this fix? Thanks

aritzolea (approved):

PR #782, opened 2025-01-20
[15.0][IMP] account_check_printing_report_base: Remove unnecessary dependency

CarlosRoca13:
`report_wkhtmltopdf_param` is not beeing used so remove it as dependency.

In version 10, this dependency was added because the `--disable-smart-shrinking` option was being used. However, starting from version 12, this option is no longer necessary, so keeping the dependency is neither required nor correct.

The dependency has been added to the modules that specifically use the flag, in this case, `account_check_printing_report_dlt103` and `account_check_printing_report_sslm102`.

cc @Tecnativa

ping @pedrobaeza @carlosdauden

### DIFF
--- a/account_check_printing_report_base/__manifest__.py
+++ b/account_check_printing_report_base/__manifest__.py
@@ -15,7 +15,7 @@
     "Odoo Community Association (OCA)",
     "category": "Generic Modules/Accounting",
     "website": "https://github.com/OCA/account-payment",
-    "depends": ["account_check_printing", "report_wkhtmltopdf_param"],
+    "depends": ["account_check_printing"],
     "external_dependencies": {"python": ["num2words"]},
     "data": [
         "security/ir.model.access.csv",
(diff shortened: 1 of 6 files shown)


pedrobaeza:
Explain a bit more the history in the commit message / PR description about initial `disable-smart-shrinking` injection and not needed anymore. The more details you always give, the better.

CarlosRoca13:
Done @pedrobaeza

pedrobaeza:
It seems the parameter still exists:

https://github.com/OCA/account-payment/actions/runs/12864018931/job/35861689362?pr=782#step:8:352

CarlosRoca13:
Yes, but not in this module, rather in `account_check_printing_report_dlt103` and `account_check_printing_report_sslm102`. Added in those modules.

pedrobaeza (approved):

carlosdauden (approved):

PR #787, opened 2025-01-28
[16.0][IMP] account_due_list: Avoid same priority than account.view_move_line_pivot

carlosdauden:
because account.action_account_moves_all have not defined view and it may be the case that the one in this module is used when it should not be.

@Tecnativa TT54575


carlosdauden:
ping @sergio-teruel @victoralmau

pedrobaeza (approved):
We can fast-track this:

/ocabot merge patch

PR #664, opened 2023-08-13
[16.0][ADD] account_payment_tier_validation

alexeirivera87:

PR #797, opened 2025-02-17
[18.0][ADD] account_payment_tier_validation

rjaraspearhead:
Add Account Payments Tier Validation


account_payment_tier_validation/views/account_payment_view.xml
```diff
@@ -0,0 +1,43 @@
...
+        <field name="model">account.payment</field>
+        <field name="inherit_id" ref="account.view_account_payment_search" />
+        <field name="arch" type="xml">
+            <filter name="reconciled" position="after">
+                <separator />
+                <filter
+                    name="needs_review"
+                    string="Needs my Review"
+                    domain="[('reviewer_ids','in',uid), ('state', 'not in', ['done', 'cancel','to approve'])]"
+                    help="My Purchases to review"
```
celm1990:
```suggestion
                    help="My Payments to review"
```

account_payment_tier_validation/views/account_payment_view.xml
```diff
@@ -0,0 +1,43 @@
...
+                />
+                <filter
+                    name="tier_validated"
+                    string="Validated"
+                    domain="[('validated', '=', True)]"
+                    help="POs validated and ready to be confirmed"
```
celm1990:
```suggestion
                    help="Payments validated and ready to be confirmed"
```

account_payment_tier_validation/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2017 ForgeFlow S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Payment Tier Validation",
+    "summary": "Extends the functionality of Payment to "
+    "support a tier validation process.",
+    "version": "18.0.1.0.0",
+    "category": "Generic Modules/Payment",
+    "website": "https://github.com/OCA/account-payment",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
```
celm1990: Why ForgeFlow?

celm1990 (changes requested):

account_payment_tier_validation/views/account_payment_view.xml
rjaraspearhead: done

account_payment_tier_validation/views/account_payment_view.xml
rjaraspearhead: done

account_payment_tier_validation/__manifest__.py
rjaraspearhead: It was a confusion when copying parts of code from other modules. I have already corrected it and changed ForgeFlow to Spearhead along with my user.

rjaraspearhead:
I have added tests to the model. Your help in validating them would be appreciated!

celm1990 (approved):
LGTM
@StefanRijnhart @bosd @pedrobaeza could you please review this? Thanks in advance!

test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo_test_helper
```
StefanRijnhart: I don't see you using this library, so I think it can be removed.

StefanRijnhart (commented):
Thanks! It's a shame https://github.com/OCA/account-payment/pull/664 never made it, but things have gotten a little more simple now that the state field is stored on the account.payment model.

StefanRijnhart:
You can simplify your commit message to `[ADD] account_payment_tier_validation`. We don't want version numbers in the commit messages.

test-requirements.txt
rjaraspearhead: I'm actually using this library for the tests. It's being used in base_tier_validation, and the tests wouldn't pass without it.

rjaraspearhead: I have removed the dependency.

StefanRijnhart (commented):
Functional test: I configure a validation tier for all payments. On a vendor bill, I click `Pay`. When I click Create Payment, the validation required error pops up. The record is not saved. This way of creating payments no longer works.

image
image

rjaraspearhead:
@StefanRijnhart 
The module will not validate the payment due to the validation configuration.

What would be the best approach in this case?

- Create the payment in draft state and allow it to go through the approval process.
- Display a message indicating that payments cannot be made this way because they require approval.

If you have any suggestions on the best way to handle this, please let me know!

StefanRijnhart:
Oh, I actually had the same issue in stock_scrap_tier_validation. I ended up changing the window action to open the new record not in a popup but in the full view. That way it could be saved without the exception. See https://github.com/OCA/stock-logistics-workflow/pull/1654

rjaraspearhead:
Thanks for the suggestion! In this case, there is now a wizard that generates the payment and performs the reconciliation. If the current flow is replaced to redirect to the payment form view, the wizard would no longer be used. What would happen then with the validations that allow processing the payment? Additionally, direct reconciliation would no longer exist.

StefanRijnhart:
Let's keep on using the wizard. Maybe you can get it to create the payments but not post them if they require validation?

celm1990:
> Functional test: I configure a validation tier for all payments. On a vendor bill, I click `Pay`. When I click Create Payment, the validation required error pops up. The record is not saved. This way of creating payments no longer works.
> 

@StefanRijnhart  I  think this use case is out of scope for this module.

The main reason is that this wizard uses `account.payment.register` and not `account.payment`, so tier validation is not applied to this model. When clicking the `Create Payment` button, the actual payment is only created at that moment, making it impossible to implement your suggestion.

image

Please add this issue to the `ROADMAP`, and let's move forward.

account_payment_tier_validation/tests/test_tier_validation.py
```diff
@@ -0,0 +1,79 @@
+# Copyright 2025 Spearhead - Ricardo Jara
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo.exceptions import ValidationError
+from odoo.tests import common, tagged
+
```
bosd, line 8: BaseCommon could be used for the test to speedup the process and prevent sending email.

from odoo.addons.base.tests.common import BaseCommon

bosd (commented):
Functional test, LGTM

StefanRijnhart:
@celm1990 Sure, I can live with a big fat disclaimer that this breaks the Pay button on invoices.

account_payment_tier_validation/tests/test_tier_validation.py
rjaraspearhead: Done

bosd (approved):

rjaraspearhead:
@StefanRijnhart 
ADD Roadmap https://github.com/OCA/account-payment/blob/824cc358c42b2ea065d932d56c26aa47b46b3e8c/account_payment_tier_validation/readme/ROADMAP.md

StefanRijnhart (approved):
/ocabot merge nobump

PR #802, opened 2025-02-24
[16.0][FIX] account_payment_term_extension: Post-install test + fallback to load CoA

pedrobaeza:
Since odoo/odoo@d0342c8, the default existing company is not getting a CoA automatically, provoking than the current tests fail with error:

```
odoo.exceptions.UserError: No journal could be found in company My Company (San Francisco) for any of those types: purchase
```

provoked by the lack of a CoA installed.

Thus, we put tests post-install for being sure localization modules are installed, the same as AccountTestInvoicingCommon does, but we don't inherit from it, as it creates an overhead creating 2 new companies and loading their CoA, and some more stuff, while we don't need all of that.

Besides, if you don't have `l10n_generic_coa` installed, you can't use another CoA (like `l10n_es`) easily, so we put little code to select the first available CoA.

@Tecnativa

### DIFF
--- a/account_payment_term_extension/tests/test_account_payment_term_multi_day.py
+++ b/account_payment_term_extension/tests/test_account_payment_term_multi_day.py
@@ -1,9 +1,10 @@
 import odoo.tests.common as common
 from odoo import fields
 from odoo.exceptions import ValidationError
-from odoo.tests.common import Form
+from odoo.tests.common import Form, tagged
 
 
+@tagged("post_install", "-at_install")
 class TestAccountPaymentTermMultiDay(common.TransactionCase):
     @classmethod
     def setUpClass(cls):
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Tested on integration tests and working:

/ocabot merge patch

account_payment_term_extension/tests/test_account_payment_term_multi_day.py
```diff
@@ -1,9 +1,10 @@
 import odoo.tests.common as common
 from odoo import fields
 from odoo.exceptions import ValidationError
-from odoo.tests.common import Form
+from odoo.tests.common import Form, tagged
 
 
+@tagged("post_install", "-at_install")
```
lmignon, line 7: Is this change required?

PR #803, opened 2025-02-25
[16.0][FIX] account_payment_return: Post-install test + fallback to load CoA

victoralmau:
Since odoo/odoo@d0342c8, the default existing company is not getting a CoA automatically, provoking than the current tests fail with the error:

`odoo.exceptions.UserError: No journal could be found in company My Company (San Francisco) for any of those types: sale`

Thus, we put tests post-install for being sure localization modules are installed, the same as AccountTestInvoicingCommon does, but we don't inherit from it, as it creates an overhead creating 2 new companies and loading their CoA and some more stuff, while we don't need all of that.

Besides, if you don't have `l10n_generic_coa` installed, you can't use another CoA (like `l10n_es`) easily, so we put little code to select the first available CoA.

@Tecnativa

### DIFF
--- a/account_payment_return_import_iso20022/wizard/payment_return_import.py
+++ b/account_payment_return_import_iso20022/wizard/payment_return_import.py
@@ -31,9 +31,12 @@ class PaymentReturnImport(models.TransientModel):
     def _parse_file(self, data_file):
         data_file_elements = self._xml_split_file(data_file)
         payment_returns = []
-        for data_file_element in data_file_elements:
-            payment_returns.extend(self._parse_single_document(data_file_element))
-        return payment_returns
+        try:
+            for data_file_element in data_file_elements:
+                payment_returns.extend(self._parse_single_document(data_file_element))
+            return payment_returns
+        except Exception:
+            return super()._parse_file(data_file)
 
     @api.model
     def _parse_single_document(self, data_file):
--- a/account_payment_return/tests/test_payment_return.py
+++ b/account_payment_return/tests/test_payment_return.py
@@ -5,9 +5,11 @@
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
 from odoo.exceptions import UserError, ValidationError
+from odoo.tests import tagged
 from odoo.tests.common import Form, TransactionCase
 
 
+@tagged("post_install", "-at_install")
 class TestPaymentReturn(TransactionCase):
     @classmethod
     def setUpClass(cls):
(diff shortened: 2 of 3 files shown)


account_payment_return_import_iso20022/wizard/payment_return_import.py
```diff
@@ -45,6 +46,12 @@ def _parse_single_document(self, data_file):
         """
         camt_parser = CamtParser()
         pain_parser = PainParser()
+        # If we are in test mode we only want this behavior when we pass that context
+        # key, that is, when we are running the tests of that module.
+        if config["test_enable"] and not self.env.context.get(
```
pedrobaeza: Why is this needed?

victoralmau: To avoid this error now that the tests are in post-install https://github.com/OCA/account-payment/actions/runs/13520670168/job/37779048967#step:8:352

pedrobaeza: I think this is a bug that the importer shouldn't fail when something different from a XML is given to import. Just don't return early and call super for following the chain.

victoralmau: Changed to apply proper fix (IMO).

pedrobaeza (commented):
This still seems incorrect. You should call `super` when the file is not a XML one.

victoralmau:
Sorry, but why block this if the fix in this related case “_Post-install test + fallback to load CoA_” is clear? 

I at least am not going to defend why this is this way https://github.com/OCA/account-payment/blob/977c1a20df419e77e81f8bffa23ed65ff905db78/account_payment_return_import_iso20022/wizard/payment_return_import.py#L48-L62 because it will be for something, and changing the behavior I BELIEVE is not the right thing to do.

pedrobaeza:
It's not changing the behavior, but fixing the behavior in the proper way. All of this have been because passing to post-install has uncovered a problem, and the problem is that the architecture, as it's though, it's not being respected in the iso20022 module, as instead of calling super when the file is not an XML to let the CSV parser to act, it's just aborting the import process. What you have done with the concatenated try/reraise is to play with the exception types to bypass it, but it's not the proper fix.

Look that this same scheme is done on the statement import, and how is handled in N43 format:

https://github.com/OCA/l10n-spain/blob/aaab375c85e33dff2708396690bf00995f69d166/l10n_es_account_statement_import_n43/wizards/account_statement_import_n43.py#L361

so what you have to do is to catch that a file is not XML (or even not the expected XSD format), and if not, call super for letting the rest of the parsers to act.

victoralmau:
The error in the tests detected was precisely because in https://github.com/OCA/account-payment/blob/977c1a20df419e77e81f8bffa23ed65ff905db78/account_payment_return_import_iso20022/wizard/camt_parser.py#L167 there was no try/except, therefore, the test failed https://github.com/OCA/account-payment/blob/977c1a20df419e77e81f8bffa23ed65ff905db78/account_payment_return_import/tests/test_import_base.py#L130 because it did not return a UserError.

With the current change it ends up calling super() in https://github.com/OCA/account-payment/blob/977c1a20df419e77e81f8bffa23ed65ff905db78/account_payment_return_import_iso20022/wizard/payment_return_import.py#L62 and for that reason UserError is returned.

The code is not pretty and/or readable, yes, but I guess, that would be another debate.

pedrobaeza:
Well, let me do it the changes I have in mind by myself and check the current test suite. Maybe that one is not correct as well.

pedrobaeza (approved):
@victoralmau you can see now what I was talking about. You have to call super, but in the proper method that is doing the iteration.

/ocabot merge patch

PR #806, opened 2025-02-27
[IMP] account_check_printing_report_base: use setUpClass for better perfomance

xaviedoanhduy:
> If there's a test class using setUp to generate test records, move it to setUpClass as it's way more performant since it runs only once.


AaronHForgeFlow (approved):
:+1:

pedrobaeza (commented):
While you are on it, removing the dependency of demo data inside tests is also advisable.

PR #807, opened 2025-02-27
[16.0][FIX] account_payment_term_extension: Define the appropriate value in some cases

victoralmau:
Apply the same behavior as in `account` https://github.com/odoo/odoo/blob/733b1474e75b32c3a467b7c3f0a83c75467d4e50/addons/account/models/account_payment_term.py#L112.

Several problems are solved

```
ERROR prod odoo.addons.account_payment_term_extension.tests.test_account_payment_term_multi_day: FAIL: TestAccountPaymentTermMultiDay.test_invoice_multi_payment_term_round Traceback (most recent call last):
  File "/opt/odoo/auto/addons/account_payment_term_extension/tests/test_account_payment_term_multi_day.py", line 322, in test_invoice_multi_payment_term_round
    self.assertEqual(
AssertionError: 500.1 != 500 : Incorrect round for invoice with payment days on 15 and 20 then 5 and 10 (round)

```
If `account_invoice_report_due_list` was installed, an error was displayed in tests

```
ERROR prod odoo.addons.account_invoice_report_due_list.tests.test_invoice_report_due_list: FAIL: TestInvoiceReportDueList.test_due_list_currency_extra Traceback (most recent call last):
  File "/opt/odoo/auto/addons/account_invoice_report_due_list/tests/test_invoice_report_due_list.py", line 123, in test_due_list_currency_extra
    self.assertEqual(invoice2.get_multi_due_list()[0][2], 50.0)
AssertionError: 32.7 != 50.0
```

@Tecnativa

### DIFF
--- a/account_payment_term_extension/models/account_payment_term.py
+++ b/account_payment_term_extension/models/account_payment_term.py
@@ -239,10 +239,10 @@ class AccountPaymentTerm(models.Model):
         untaxed_amount_left = untaxed_amount
         untaxed_amount_currency_left = untaxed_amount_currency
 
-        total_amount = remaining_amount = tax_amount + untaxed_amount
-        total_amount_currency = remaining_amount_currency = (
-            tax_amount_currency + untaxed_amount_currency
-        )
+        total_amount = tax_amount + untaxed_amount
+        total_amount_currency = tax_amount_currency + untaxed_amount_currency
+        foreign_rounding_amount = 0
+        company_rounding_amount = 0
         result = []
         precision_digits = currency.decimal_places
         company_precision_digits = company_currency.decimal_places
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
Can we avoid overwriting the full method and instead just touch where needed?

victoralmau:
> Can we avoid overwriting the full method and instead just touch where needed?

But the method should do the same as the standard does, right? Since the method was “_copied_” until now it has changed, that's why the data is incorrect. 
Maybe we can create a PR to odoo so we can extend the method instead of having to overwrite it completely?

pedrobaeza (approved):
Yes, but I don't have faith in having it merged. That's why I tell you to try to do it here. Anyway, let's merge for now for having this correct.

/ocabot merge patch

PR #809, opened 2025-03-12
[16.0][ADD] payment_plaid

crrodrigueztrujillo:
@BinhexTeam 
Integration with plaid


payment_plaid/README.rst
```diff
@@ -0,0 +1,64 @@
+.. image:: https://github.com/OCA/maintainer-tools/blob/master/template/module/icon.png
+:alt: Odoo Community Association
+:target: https://odoo-community.org
+
+================
+Payment Plaid
+
+Integración de Odoo con Plaid para facilitar pagos bancarios y transferencias.
```
rvalyi: there should be a standard OCA readme folder instead. And as this addon is not in a specific localization repo, I guess the doc should be in English then.

rvalyi (changes requested):

payment_plaid/README.rst
crrodrigueztrujillo: @rvalyi  the file has already been updated

PR #840, opened 2025-06-03
[18.0][ADD] account_payment_method_base

alexis-via:
This PR replaces the PR https://github.com/OCA/account-financial-tools/pull/2094 following the suggestion of Pedro that this project would be more suitable.
I took into account the remarks made on my previous PR:
- remove active field
- add nocreate="1" on list view
- improve _compute_display_name to have a translated value for payment_type


rvalyi (approved):
approved as I tested before https://github.com/OCA/account-financial-tools/pull/2094#pullrequestreview-2880186747

@alexis-via as for the line_ids o2m field I trust your analysis, I pointed it in case you just forgot it.

sbidoul (approved):
Also good for me, with comments from the https://github.com/OCA/account-financial-tools/pull/2094 addressed.

alexis-via:
@rvalyi for the moment, I prefer not to display the lines on the form view. If we find it useful in the end, we can still improve the module later.

PR #842, opened 2025-06-04
[18.0][IMP] account_payment_method_base: move ir.model.access.csv from account_payment_base_oca

alexis-via:


alexis-via:
@sbidoul

PR #525, opened 2022-07-26
[FIX] account_payment_return: partial_reconcile_returned_ids is kept…

carlosdauden:
… when invoice is duplicated

@Tecnativa TT38126

PR #849, opened 2025-07-01
[16.0][FIX] account_payment_return: partial_reconcile_returned_ids is kept when invoice is duplicated

victoralmau:
FWP from 13.0: https://github.com/OCA/account-payment/pull/525

partial_reconcile_returned_ids is kept when invoice is duplicated

@Tecnativa TT38126

### DIFF
--- a/account_payment_return/models/account_move.py
+++ b/account_payment_return/models/account_move.py
@@ -63,6 +63,7 @@ class AccountMoveLine(models.Model):
         relation="account_partial_reconcile_account_move_line_rel",
         column1="move_line_id",
         column2="partial_reconcile_id",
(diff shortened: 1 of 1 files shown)


pedrobaeza:
keeped > kept

pedrobaeza:
You need the same in https://github.com/OCA/account-payment/pull/849/files#diff-6f60c5be3a122ae8debe3ca75d51102d7fe0a89cba74fc7b8a264e1b189eac71R73, although you don't usually duplicate that record, but technically it can happen the same problem.

pedrobaeza (approved):

carlosdauden (approved):

PR #858, opened 2025-07-17
[MIG] account_payment_notification: v18

yajo:
Behavior changes:
- Now, notifications are put in outgoing queues only if done massively. This is the standard behavior in v18.

Test changes and improvements:
- Now they are executed with lower permissions.
- We make sure the company is consistent accross the whole test.
- SMS bodies are now compared using HTML assertions.
- Messages no longer have a `description` field.

@moduon MT-10470


loida-vm:
@yajo As you say:

_Now, notifications are put in outgoing queues only if done massively. This is the standard behavior in v18._
So, is this correct?
As you can see, a notification was sent to a single partner, and then massively. Both are in the outgoing queue.
image

yajo:
Hmm... how did you test that? It's not happening to me. I even tested in runboat:


https://github.com/user-attachments/assets/5ff8f8a2-776c-4b78-be5e-aecdf5c4e7ab

loida-vm:
Here in the video I explain to you the different cases, because we are creating the payments by different methods. 
https://www.loom.com/share/41f87ddffc9647a2ac1e4125aaa62c07?sid=b8933ef4-e764-4894-8c0f-dfc4edb37f0a
I found different bugs

yajo:
Thanks for the in-depht review!

### Green vs gray bakgrounds

When you are in a form and want to send the notification, you see a wizard before posting the notification message. With the wizard, you would be able to change the contents of the notification before posting them. That makes Odoo think it's a human-made message, and that's why it appears green.

Other messages, those that are completely automated, have no human interaction. That's why Odoo renders them gray.

In any case, the envelope will tell you who was notified, no matter if green or gray. So, although surprising, I don't think that's a bug.

### Mail queued when notifying a single payment from the list view

Thanks I didn't notice that. It seems then that the mail is sent directly when interacting directly from the form view, but is always queued when using the list view.

IMHO it doesn't seem a big problem as long as it's properly documented, so I'll update the README.

### Notifications not sent when paid from the invoice

~~Indeed this is a bug. I'll fix it.~~ Explained in comment below.

loida-vm:
Thanks, we’ll wait for the changes to review the README, as well as the fix for the bug.

yajo:
> Notifications not sent when paid from the invoice

It turns out this was working as expected.

The module notifies when payments are marked as *sent*, not as *paid*. However, when paying from an invoice, it is automatically marked as *paid*, not as *sent*.

I have added a few details to README. In summary, it's working as expected.

yajo:
The CI is ❌ because of a flaky test in another module, which I'm fixing in https://github.com/OCA/account-payment/pull/865. It has no relationship with this module.

loida-vm (approved):
LGTM! The module works. 
Thanks for making it clear in the README @yajo

yajo:
@OCA/banking-maintainers this PR is ready, could someone merge please? Thanks!

rafaelbn (approved):
Tested funcionally 👍🏼 

We are the owners and maintainers so we should be able to merge our selves ❓

PR #860, opened 2025-07-18
[18.0][OU-FIX] account_check_printing_report_base: column exists

FernandoRomera:
In Odoo:
[[    bank_check_printing_layout = fields.Selection(](https://github.com/odoo/odoo/blob/c5222a82887b6c5c781fe823865e2573c14e8b7b/addons/account_check_printing/models/account_journal.py#L37)](https://github.com/odoo/odoo/blob/c5222a82887b6c5c781fe823865e2573c14e8b7b/addons/account_check_printing/models/account_journal.py#L37C5-L40C6)

@CarlosRoca13 
Can you check if the modification is timely or does not make sense?.

### DIFF
--- a/account_check_printing_report_base/migrations/18.0.1.0.0/pre-migration.py
+++ b/account_check_printing_report_base/migrations/18.0.1.0.0/pre-migration.py
@@ -6,14 +6,10 @@ from openupgradelib import openupgrade
 
 @openupgrade.migrate()
 def migrate(env, version):
-    openupgrade.rename_fields(
-        env,
-        [
-            (
-                "account.journal",
-                "account_journal",
-                "account_check_printing_layout",
-                "bank_check_printing_layout",
-            ),
-        ],
+    openupgrade.logged_query(
+        env.cr,
+        """
+            UPDATE account_journal
+            SET bank_check_printing_layout = account_check_printing_layout
+            """,
     )


FernandoRomera:
@CarlosRoca13 
Sorry, one more question, the column has to be deleted?.

account_check_printing_report_base/migrations/18.0.1.0.0/pre-migration.py
```diff
@@ -6,14 +6,13 @@
...
-                "account.journal",
-                "account_journal",
-                "account_check_printing_layout",
-                "bank_check_printing_layout",
-            ),
-        ],
-    )
+    if openupgrade.column_exists(
+        env.cr, "account_journal", "account_check_printing_layout"
+    ):
```
CarlosRoca13:
```suggestion
```
I don't think this is necessary, because the module already depends on `account_check_printing`, so the data will already exist. ping @pedrobaeza

CarlosRoca13 (commented):
True :+1: We need to keep the data in case any issues arise after the migration so don't delete the column

pedrobaeza (changes requested):
AFAIK, this scripts is only executed if you come from 17.0, with both modules this one and `account_check_printing` from Odoo installed, so the column should exist, and the current code is correct. Can you describe the steps to reproduce the problem?

account_check_printing_report_base/migrations/18.0.1.0.0/pre-migration.py
FernandoRomera: Ok.
Remove the code, thanks

pedrobaeza:
Yes, but I still don't understand the current change.

FernandoRomera:
@pedrobaeza 
In 17.0, the table account_journal have the column: 
account_check_printing_layout

when upgrade the module "account_check_printing", add the column :
bank_check_printing_layout

now the table account_journal, have the 2 columns:
account_check_printing_layout and bank_check_printing_layout

and when upgrade the module "account_check_printing_report_base", the premigration do:
alter table account_journal rename account_check_printing_layout to bank_check_printing_layout and throw an error.

So I think it is better to do an update and avoid the error.

pedrobaeza (approved):
OK, understood.

FernandoRomera:
@pedrobaeza 
Excuse me, one question, do you think it would be convenient to rename the column “account_check_printing_layout” of the table “res.company” to “bank_check_printing_layout” (to normalize). If so, I can take care of making a PR.

pedrobaeza:
Well, I think the current one is OK. Why do you think renaming is better?

pedrobaeza:
Oh, wait, you are talking about `res.company`, not `account.journal`. If you think there's a perfect equivalency, then OK, and it should be done in the proper module.

FernandoRomera:
When you indicate the appropriate module, you mean the “account_check_printing_report_base” module.

pedrobaeza:
Well, I don't know right now which one creates it...

PR #848, opened 2025-06-24
[18.0][MIG] account_financial_discount: Migrated force feature to 18.0

grindtildeath:

PR #863, opened 2025-07-30
[18.0][MIG] account_financial_discount: Migrated force feature to 18.0

grindtildeath:
Supersedes: #848


account_force_early_discount/models/account_move.py
```diff
@@ -0,0 +1,72 @@
...
+    @api.depends(
+        "invoice_payment_term_id",
+        "invoice_payment_term_id.early_discount",
+        "force_early_discount",
+        "state",
+    )
+    def _compute_display_force_early_discount(self):
+        """Compute discount financial discount fields"""
+        for rec in self:
+            rec.display_force_early_discount = rec._get_display_force_early_discount()
```
ivantodorovich: I'd just get rid of `_get_display_force_early_discount` and compute everything here.
IMO it makes it easy to see the fields dependencies

account_force_early_discount/models/account_move.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def _get_first_payment_term_line(self):
+        self.ensure_one()
+        payment_term_lines = self.line_ids.filtered(
+            lambda line: line.display_type == "payment_term"
+        )
+        return fields.first(payment_term_lines.sorted("date_maturity"))
```
ivantodorovich: For a long time I also used `fields.first`.
But then I realized, by reading its docstring, that it really doesn't make sense to use it in most cases.

```
def first(records):
    """ Return the first record in ``records``, with the same prefetching. """
    return next(iter(records)) if len(records) > 1 else records
```

Essentially, you get the first record with the prefetching set to all records.. but you usually don't really care about the other records so why would you want them prefetched?

Instead, the pythonic way works well and better fits most usecases:

```suggestion
        return payment_term_lines.sorted("date_maturity")[:1]
```

That'll give you just the first record, with no prefetching on the others

account_force_early_discount/wizard/account_payment_register.py
```diff
@@ -0,0 +1,110 @@
...
+          with force early discount.
+        - Early payment discount is available on any invoice and wizard is marked
+          with force early discount."""
+        for wizard in self:
+            if wizard.force_early_discount:
+                wizard.with_forced_early_discount = True
+                continue
+            lines_with_forced_discount_past_date = wizard.line_ids.filtered(
+                lambda li: li.discount_amount_currency
+                and li.discount_date < wizard.payment_date  # noqa: B023
```
ivantodorovich, line 38: AFAICS `account.move.line.discount_date` is not required, so it can potentially be `False`.
That can lead to 

```
TypeError: '<' not supported between instances of 'bool' and 'datetime.datetime'
```

account_force_early_discount/wizard/account_payment_register.py
```diff
@@ -0,0 +1,110 @@
...
+                lines_with_forced_discount_past_date
+            )
+
+    @api.depends("line_ids")
+    def _compute_show_force_early_discount(self):
+        for wizard in self:
+            # If at least one invoice has late discounts on its move lines
+            wizard.show_force_early_discount = wizard.line_ids.filtered(
+                lambda li: li.discount_amount_currency
+                and li.discount_date < wizard.payment_date  # noqa: B023
```
ivantodorovich, line 51: same

account_force_early_discount/wizard/account_payment_register.py
ivantodorovich: Moreover, shouldn't it be lower-or-equal, instead?
If the customer pays on the very last day, I guess the discount is still valid

account_force_early_discount/models/account_move.py
grindtildeath: Goal was to keep the diff as small as possible, but I agree we don't need this function anymore

account_force_early_discount/models/account_move.py
grindtildeath: AFAIR we used to recommend using `fields.first` to avoid an error when slicing using `[0]`, but I'm not sure why we never came up with this before :confused:

account_force_early_discount/wizard/account_payment_register.py
grindtildeath: Actually, I guess the `TypeError` will not pop up due to the first condition on `discount_amount_currency`, as there shouldn't be a line with a discount amount and no discount date.

Then yes, the discount is valid until the last day, but that's already handled by the standard. As this module is about allowing to force discount after the date is passed, I guess we don't care about those that are paid in time, and the variable being called "past date" seems to confirm that.

account_force_early_discount/models/account_move.py
```diff
@@ -0,0 +1,86 @@
...
+        """Compute discount financial discount fields"""
+        for rec in self:
+            display_force = False
+            if rec.state == "draft":
+                if (
+                    rec.invoice_payment_term_id.discount_days
+                    and rec.invoice_payment_term_id.discount_percentage
+                ):
+                    display_force = True
+            elif self.state == "posted":
```
simahawk:
```suggestion
            elif rec.state == "posted":
```
:fire:

account_force_early_discount/tests/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import common
```
simahawk:
```suggestion
```

account_force_early_discount/tests/common.py
```diff
@@ -0,0 +1,95 @@
...
+                line_form.name = product and product.name or "test"
+                line_form.quantity = quantity
+                line_form.price_unit = unit_price
+                if tax:
+                    line_form.tax_ids = tax
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
```
simahawk: pls set tracking disable

account_force_early_discount/tests/test_financial_discount_manual_payment.py
```diff
@@ -0,0 +1,389 @@
+# Copyright 2019-2020 Camptocamp SA
```
simahawk, line 1:
```suggestion
# Copyright 2019 Camptocamp SA
```

account_force_early_discount/models/account_move.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def _get_first_payment_term_line(self):
+        self.ensure_one()
+        payment_term_lines = self.line_ids.filtered(
+            lambda line: line.display_type == "payment_term"
+        )
+        return payment_term_lines.sorted("date_maturity")[:1]
+
+    def _is_eligible_for_early_payment_discount(self, currency, reference_date):
+        force_move_ids = self.env.context.get("_force_early_discount")
```
simahawk:
```suggestion
        force_move_ids = self.env.context.get("account_force_early_discount_move_ids")
```

account_force_early_discount/wizard/account_payment_register.py
```diff
@@ -0,0 +1,110 @@
...
+        batch_result = self._apply_force_early_payment_discount_ctx(batch_result)
+        return super()._create_payment_vals_from_batch(batch_result)
+
+    def _get_total_amounts_to_pay(self, batch_results):
+        # Set ctx to be used on account.move._is_eligible_for_early_payment_discount
+        if self.with_forced_early_discount:
+            force_ids = []
+            for batch_result in batch_results:
+                if self.force_early_discount:
+                    batch_result["lines"] = batch_result["lines"].with_context(
```
simahawk, line 95: shouldn't be using the helper method above for the ctx?

simahawk (changes requested):

account_force_early_discount/models/account_move.py
grindtildeath: Good catch :+1:

account_force_early_discount/tests/test_financial_discount_manual_payment.py
grindtildeath: Should we really change what's existing there?

account_force_early_discount/wizard/account_payment_register.py
grindtildeath: IIRC I did not use the helper because we still need to collect all move ids to force to alter the context on self. IMO it doesn't make the code easier to understand if we have to loop twice over the batch results :shrug:

ivantodorovich (approved):

account_force_early_discount/tests/test_financial_discount_manual_payment.py
simahawk: if it's wrong, yes

simahawk (approved):
to be squashed, pre-approving

account_force_early_discount/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Account force early payment discount",
+    "summary": "Allow forcing financial discounts for early payments",
+    "version": "18.0.1.0.1",
```
TDu:
```suggestion
    "version": "18.0.1.0.0",
```

account_force_early_discount/models/account_move.py
```diff
@@ -0,0 +1,72 @@
...
+    def _get_first_payment_term_line(self):
+        self.ensure_one()
+        payment_term_lines = self.line_ids.filtered(
+            lambda line: line.display_type == "payment_term"
+        )
+        return payment_term_lines.sorted("date_maturity")[:1]
+
+    def _is_eligible_for_early_payment_discount(self, currency, reference_date):
+        force_move_ids = self.env.context.get("_force_early_discount_move_ids")
+        if self.force_early_discount or force_move_ids:
```
TDu:
```suggestion
        if self.force_early_discount or self.id in force_move_ids:
```
And remove the condition below

account_force_early_discount/models/account_move.py
```diff
@@ -0,0 +1,72 @@
...
+            payment_terms = self.line_ids.filtered(
+                lambda line: line.display_type == "payment_term"
+            )
+            return (
+                (self.force_early_discount or self.id in force_move_ids)
```
TDu:
```suggestion
           
```

account_force_early_discount/models/account_move.py
```diff
@@ -0,0 +1,72 @@
...
+                and self.currency_id == currency
+                and self.move_type
+                in ("out_invoice", "out_receipt", "in_invoice", "in_receipt")
```
TDu:
```suggestion
                in self._early_payment_discount_move_types()
```

TDu (approved):

TDu:
@grindtildeath Could you squash and see if the pre-commit still have an issue, thanks

TDu:
@grindtildeath CI is red :crying_cat_face:

TDu:
Thanks @grindtildeath  CI is :green_heart:  again
Squash and merge next ?

PR #868, opened 2025-08-06
[16.0][FIX] account_cash_invoice: Journal selectable

pedrobaeza:
The journal where to pay should be selectable at user level, so the readonly flag is not correct.

Initializing also with the active_id is not correct, as you may come from other screens like PoS session.

Finally, the field should be restricted to proper journals, not all.

@Tecnativa TT57569

### DIFF
--- a/account_cash_invoice/wizard/cash_pay_invoice.py
+++ b/account_cash_invoice/wizard/cash_pay_invoice.py
@@ -14,16 +14,6 @@ class CashPayInvoice(models.TransientModel):
     _name = "cash.pay.invoice"
     _description = "Cash Pay invoice from bank statement"
 
-    def _default_company(self):
-        active_ids = self.env.context.get("active_ids")
-        journal = self.env["account.journal"].browse(active_ids)
-        return journal.company_id
-
-    def _default_currency(self):
-        active_ids = self.env.context.get("active_ids")
-        journal = self.env["account.journal"].browse(active_ids)
-        return journal.currency_id or journal.company_id.currency_id
-
     invoice_id = fields.Many2one(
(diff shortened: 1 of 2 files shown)


CarlosRoca13 (approved):

victoralmau (approved):

account_cash_invoice/wizard/cash_pay_invoice.py
```diff
@@ -34,22 +24,15 @@ def _default_currency(self):
...
-        default=lambda self: self._default_currency(),
-        required=True,
-        readonly=True,
+        comodel_name="res.currency", compute="_compute_currency_id"
     )
     journal_id = fields.Many2one(
         comodel_name="account.journal",
         required=True,
-        readonly=True,
-        default=lambda self: self.env.context.get("active_ids")[0],
```
carlos-lopez-tecnativa, line 52: Did you take into account whether it’s necessary to update the code in this module? The intention is to pass the `cash_journal_id` from the pos.session.

https://github.com/OCA/pos/blob/a6b2d9e52bcb02c92c1ebdd6ac88d6bba6cc26b8/pos_session_pay_invoice/models/pos_session.py#L30-L35

carlos-lopez-tecnativa (commented):
Sorry for arriving late to the party.

account_cash_invoice/wizard/cash_pay_invoice.py
pedrobaeza: I don't see any drawback right now, and the proper way to pass that is using the context key `default_journal_id`, not that way.

PR #793, opened 2025-02-06
[18.0][MIG] account_payment_return_import: Migration to 18.0

absal-smile:
## Dependencies

This PR depends on:
- [x] PR #792: account_payment_return (https://github.com/OCA/account-payment/pull/792)

PR #870, opened 2025-08-18
[18.0][MIG] account_payment_return_import: Migration to 18.0

victoralmau:
Migration to 18.0

Superseed https://github.com/OCA/account-payment/pull/793

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT54092


account_payment_return/models/payment_return_reason.py
```diff
@@ -11,7 +11,7 @@ class PaymentReturnReason(models.Model):
 
     code = fields.Char()
     name = fields.Char(string="Reason", translate=True)
-    display_name = fields.Char(compute="_compute_display_name")
+    display_name = fields.Char(compute="_compute_display_name", store=True)
```
pedrobaeza: This is not correct, as the name is translatable, and thus, you will store the result in only one language. Which is the reason for this?

victoralmau: This is to avoid the error log:

```
ERROR devel odoo.osv.expression: Non-stored field payment.return.reason.display_name cannot be searched.
NoneType: None
```

pedrobaeza: Then add the `search` method instead.

account_payment_return/models/payment_return_reason.py
```diff
@@ -11,9 +12,22 @@ class PaymentReturnReason(models.Model):
...
+
+    @api.model
+    def _search_display_name(self, operator, value):
+        domain = []
+        for f_name in self._rec_names_search:
+            domain_item = [(f_name, operator, value)]
+            if domain == []:
+                domain = domain_item
+            else:
+                domain = expression.OR([domain, domain_item])
```
pedrobaeza: The join should be `AND` if the operator is in negative operator list (there's a constant list that you can import to check its inclusion).

pedrobaeza (approved):
Merging counting with the approval in the superseded PR.

/ocabot merge nobump

PR #321, opened 2020-03-24
[12.0][IMP] - support Camt 053 in account_payment_return_import_iso20022

sbejaoui:

PR #874, opened 2025-09-01
[18.0][IMP] - support Camt 053 in account_payment_return_import_iso20022

qgroulard:
Forward port of https://github.com/OCA/account-payment/pull/321

5 years later, our customer still receives Camt 053 files for the payment returns from the bank.
It's a small addition to `account_payment_return_import_iso20022` to support them.


account_payment_return_import_iso20022/wizard/camt_parser.py
```diff
@@ -143,7 +152,7 @@ def parse_payment_returns(self, ns, node):
 
     def check_version(self, ns, root):
```
yvaucher, line 153: The error message "no camt.054.001.02:" do not take into account the compat with camt.053.

yvaucher (approved):
Just a nitpicking on the error message in `check_version` that doesn't mention camt 053 and can be confusing for the user.

Otherwise LGTM, already approving.

account_payment_return_import_iso20022/wizard/camt_parser.py
qgroulard: Fixed

yvaucher (approved):

PR #875, opened 2025-09-02
[18.0][FIX] account_payment_return_import: payment_returns domain typo

qgroulard:


pedrobaeza (approved):
Indeed! Let's fast-track it for fixing the problem.

/ocabot merge patch

PR #879, opened 2025-09-15
[18.0][IMP] account_payment_return: notify invoice of a payment return

glitchov:
Migration of #310 : for better extensibility, notify the invoice that a payment return has arrived and let it do what it implies for itself.

### DIFF
--- a/account_payment_return/models/payment_return.py
+++ b/account_payment_return/models/payment_return.py
@@ -192,6 +192,7 @@ class PaymentReturn(models.Model):
                 # payment_aml: credit on customer account (from payment move)
                 # invoice_amls: debit on customer account (from invoice move)
                 invoice_amls = payment_aml.matched_debit_ids.mapped("debit_move_id")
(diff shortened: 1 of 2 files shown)


account_payment_return/models/payment_return.py
```diff
@@ -191,6 +191,7 @@ def action_confirm(self):
                 # payment_aml: credit on customer account (from payment move)
                 # invoice_amls: debit on customer account (from invoice move)
                 invoice_amls = payment_aml.matched_debit_ids.mapped("debit_move_id")
+                invoice_amls._payment_returned(return_line)
```
qgroulard, line 195: In version 12, the line 227 was removed (`invoices.write(self._prepare_invoice_returned_vals()`). If you don't do that, the invoices are written twice.

qgroulard (changes requested):

qgroulard (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #882, opened 2025-09-23
[18.0][FIX] account_payment_return: 'memo' field in account.payment

LauraCForgeFlow:
Model `account.payment` does not have a `ref` field anymore, it has been substituted by `memo`. 

Since `line_id` in `prepare_values_returned_widget` can be either an `account.payment` or an `account.move.line` instance, we check whether we need to use the `memo` or the `ref` field.

### DIFF
--- a/account_payment_return/models/account_move.py
+++ b/account_payment_return/models/account_move.py
@@ -33,6 +33,7 @@ class AccountMove(models.Model):
             payment_method_name = line_id.payment_method_line_id.name
         except AttributeError:
             payment_method_name = False
+        ref = line_id.memo if hasattr(line_id, "memo") else line_id.ref
         return {
             "name": line_id.name,
             "journal_name": line_id.journal_id.name,
@@ -49,7 +50,7 @@ class AccountMove(models.Model):
                 currency_obj=line_id.currency_id,
             ),
             "payment_method_name": payment_method_name,
-            "ref": f"{line_id.move_id.name} ({line_id.ref})",
+            "ref": f"{line_id.move_id.name} ({ref})",
             "returned": is_return,
         }
 


AlexPForgeFlow (approved):

GuillemCForgeFlow (approved):
lgtm 👍🏿

AaronHForgeFlow (approved):
:+1:

account_payment_return/models/account_move.py
```diff
@@ -33,6 +33,7 @@ def prepare_values_returned_widget(self, line_id, amount, is_return=False):
             payment_method_name = line_id.payment_method_line_id.name
         except AttributeError:
             payment_method_name = False
+        ref = line_id.memo if hasattr(line_id, "memo") else line_id.ref
```
MiquelRForgeFlow, line 36:
```suggestion
        ref = line_id.memo if hasattr(line_id, "memo") and line_id.memo else line_id.ref
```
If `memo` is empty, let's fill with `ref`.

LauraCForgeFlow: I don't think that will work, since if `line_id` has the `memo` field (so `hasattr(line_id, "memo")` returns True), the record won't have the `ref` field. If we try to fill it with `ref` we might get an error.

pedrobaeza (commented):
AFAIK `memo` is a multi-line field, so it can't be used.

LauraCForgeFlow:
@pedrobaeza `memo` in (at least in the `account.payment` model) is a char field, so the same field type as `ref` for `account.move.line`, see [source](https://github.com/odoo/odoo/blob/cbb55390f230730d365350397580ee72014d4f63/addons/account/models/account_payment.py#L108).

account_payment_return/models/account_move.py
MiquelRForgeFlow: > if `line_id` has the `memo` field, the record won't have the `ref` field

Why?

> If we try to fill it with `ref` we might get an error.

Why?


(I don't know, I am just asking to know it better)

pedrobaeza (approved):
/ocabot merge patch

account_payment_return/models/account_move.py
pedrobaeza: Check the initial comment in the PR with the explanation: https://github.com/OCA/account-payment/pull/882#issue-3444296421

LauraCForgeFlow: Because the `line_id` parameter in this method can be, either a record of the `account.payment` model, or a record of the `account.move.line` model.

If `line_id` is an `account.payment`, then it will have the `memo` field, and id won't have the `ref` field. That's why trying to get the `ref` field in that case will throw an error, since the attribute does not exist for that record or model.

However, if `line_id` is an `account.move.line`, then it won't have the `memo` field, but it will have `ref`. So that covers the other scenario.

Hope it is clear now!! :)

PR #764, opened 2024-10-07
[16.0][FIX] account_payment_return_import_iso20022: Amount is not computed …

carlosdauden:
…when match by payment reference

This commit call supper method without processed lines in self:
https://github.com/OCA/account-payment/commit/cd89ad977665da8d005549ffcf899fc2758a3388
and those proceed lines are not computed here:
https://github.com/OCA/account-payment/blob/f6f205acd14a474074e4cd301dad8ed426bcd15c/account_payment_return/models/payment_return.py#L393 TT51075

PR #884, opened 2025-09-24
[18.0][IMP] account_payment_return_import_iso20022: Match compute amount

qgroulard:
Amounts are computed for lines matched in the super() method, but the super() is not called for lines matched from payment orders.
Thus, we add the amount compute for these lines.

### DIFF
--- a/account_payment_return_import_iso20022/models/payment_return.py
+++ b/account_payment_return_import_iso20022/models/payment_return.py
@@ -30,4 +30,5 @@ class PaymentReturnLine(models.Model):
                         == payment.destination_account_id
                         and x.partner_id == payment.partner_id
                     )
+        matched.filtered(lambda x: not x.amount)._compute_amount()
         return super(PaymentReturnLine, self - matched)._find_match()


pedrobaeza:
I don't see how is this related to payment orders. Can you put the complete steps to reproduce the problem?

qgroulard:
I am just saying this is payment orders related because this is the use case of this module according to the docstring: https://github.com/OCA/account-payment/blob/18.0/account_payment_return_import_iso20022/models/payment_return.py#L13

The point is that in the base method we compute the amount: https://github.com/OCA/account-payment/blob/18.0/account_payment_return/models/payment_return.py#L395.
And in the module `account_payment_return_import_iso20022` we don't call `super()` on all records, thus some of them don't have the amount computed.

pedrobaeza:
And shouldn't be the fix to have the computed method auto-trigger touching the dependencies? Just wondering...

@carlosdauden can you give your insights?

pedrobaeza:
@qgroulard now that I have faced the same issue, I understand the problem, but do you know what? It was already fixed in 16.0 by my colleague @carlosdauden in #764, calling `super` for the whole recordset.

If you don't mind, I'm closing this one and cherry-picking the fix from 16 instead.

pedrobaeza:
Fixes for 17 and 18 done in #886 and #887

PR #887, opened 2025-10-04
[17.0][FIX] account_payment_return_import_iso20022: Amount is not computed when match by payment reference

pedrobaeza:
Forward-port of #764

This commit call supper method without processed lines in self: https://github.com/OCA/account-payment/commit/cd89ad977665da8d005549ffcf899fc2758a3388 and that proceed lines are not computed here:
https://github.com/OCA/account-payment/blob/f6f205acd14a474074e4cd301dad8ed426bcd15c/account_payment_return/models/payment_return.py#L393

@Tecnativa TT51075


pedrobaeza:
Fast-tracking as the bug is confirmed and fixed in 16.0

/ocabot merge patch

PR #888, opened 2025-10-06
[17.0][MIG] account_check_date: Migration to 17.0

lperez-solvos:
Standard Migration to 17.0.

Changes:

- Removed deprecated Python field attributes (states, attrs).
- Updated XML views with readonly/invisible expressions.
- Adjusted post-init hook.


account_check_date/__manifest__.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2020 Ecosoft Co., Ltd. (http://ecosoft.co.th)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Account Check Date",
+    "version": "17.0.0.1.0",
```
DantePereyra:
```suggestion
    "version": "17.0.1.0.0",
```
Small detail

account_check_date/models/account_payment.py
```diff
@@ -0,0 +1,15 @@
...
+from odoo import fields, models
+
+
+class AccountPayment(models.Model):
+    _inherit = "account.payment"
+
+    check_date = fields.Date(
+        default=fields.Date.context_today,
+        required=True,
+        readonly=True,
```
DantePereyra: Could be `readonly` be on `account_payment_views.xml`?
- _If a field is marked as readonly=True on the model, it won't be possible to import it through the Odoo import tool, so avoid it as possible and define the readonly attribute on the views instead._ See more [here](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0)

DantePereyra (changes requested):
Minor comments :+1:

account_check_date/models/account_payment.py
dalonsod: Yes, this is actually unneeded. In original code `states` and `readonly` combined at field definition actually meant this:
* By default, field is readonly.
* But, for one state (`draft`), is not.

So removing `states` at python definition should be done along with `readonly` deletion. And `readonly` management 100% moved to view (that is currently done)

DantePereyra (approved):

ChristianSantamaria (approved):
LGTM

PR #894, opened 2025-10-23
[16.0][FIX] account_payment_term_partner_holiday: no recompute terms

LauraCForgeFlow:
In versions 16.0+ it is not necessary to recompute the payment terms upon posting an account move since the original `action_post` method already manages the syncing of some dynamic info, including the payment terms. 

That is also demonstrated by the `test_partner_2_invoice_change_due_date_on_confirm` test, which still passes when not recomputing the payment terms manually in this module.

### DIFF
--- a/account_payment_term_partner_holiday/models/account_move.py
+++ b/account_payment_term_partner_holiday/models/account_move.py
@@ -7,12 +7,6 @@ from odoo import api, models
 class AccountMove(models.Model):
     _inherit = "account.move"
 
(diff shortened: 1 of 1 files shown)


JordiMForgeFlow (approved):

AaronHForgeFlow (approved):
LGTM

victoralmau (approved):
The `_recompute_payment_terms_lines()` method existed in v15 and no longer exists in v16, therefore it is consistent.

Issue #843, opened 2025-06-08
Print invoice with payment return


agorbit:
Error report when you try print an invoice with a payment return 

## Module
account_payment_return.

## Describe the bug
odoo.addons.base.models.ir_qweb.QWebException: Error while render the template
KeyError: 'is_exchange'
Template: account.report_invoice_document
Path: /t/t/div[2]/div/div[4]/div[1]/div[1]/div/table/t[3]/t/t[2]/tr
Node: <tr t-if="payment_vals[\'is_exchange\'] == 0"/>

## To Reproduce
**Odoo Enterprise Edition 17**:

1. Create an invoice
2. Make a payment
3. Make a payment return
4. Try to print the standard invoice report

If you use the stardard report without payment no problem.

PR #767, opened 2024-10-29
[16.0][FIX] account_payment_return: Add 'is_exchange' value to dict returned in prepare_values_returned_widget method

sergiobstoj:
**Current behaviour:**

- When having an invoice with returns, there is an error when printing PDF:

[Grabación de pantalla desde 29-10-24 12:26:33.webm](https://github.com/user-attachments/assets/9e346ce8-75f0-4c53-a742-09f42f1dea28)

**Behaviour with this PR:**

- The PDF is generated corrrectly.

[Grabación de pantalla desde 29-10-24 12:29:16.webm](https://github.com/user-attachments/assets/e2753540-c372-4ddd-abcd-2be2e2443094)

PR #896, opened 2025-10-24
[17.0][FIX] account_payment_return: Fix error is_exchange when printing invoice with payment return

eticco:
Applied to 17.0 same fix of 16.0: https://github.com/OCA/account-payment/pull/767/commits/3fd5e6aa6f3ac487e1e9ea7d6aa5b80ac555fbf4

Fix this issue: https://github.com/OCA/account-payment/issues/843


pilarvargas-tecnativa (approved):

pilarvargas-tecnativa (changes requested):
Do not define a new commit, simply use fwport to respect the authorship of the commit.

account_payment_return/__manifest__.py
```diff
@@ -11,7 +11,7 @@
 
 {
     "name": "Account Payment Returns",
-    "version": "17.0.1.2.2",
+    "version": "17.0.1.2.3",
```
pilarvargas-tecnativa, line 14: It is not necessary to change the manifest version.

PR #897, opened 2025-10-24
[18.0][FIX] account_payment_return: Select proper outstanding account

pedrobaeza:
Odoo has changed the mechanism of selecting outstanding account: on v17 there were a field "account_journal_payment_debit_account_id" on res.company for holding the outstanding account. Now, this outstanding account is reflected through an XML-ID identifier.

Thus, the adaptation to the module should follow the same mechanism for selecting the outstanding account.

There's also no need of a specific error message, as the method already contains one.

@Tecnativa TT58856

### DIFF
--- a/account_payment_return/models/payment_return.py
+++ b/account_payment_return/models/payment_return.py
@@ -153,12 +153,13 @@ class PaymentReturn(models.Model):
 
     def _prepare_move_line(self, move, total_amount):
         self.ensure_one()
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Linked to https://github.com/OCA/OpenUpgrade/pull/5425

carlosdauden (approved):

PR #901, opened 2025-11-05
[18.0][ADD] payment_sequra: new SeQura payment provider and payment method

juancarlosonate-tecnativa:
@tecnativa TT58926
@carlos-lopez-tecnativa @pedrobaeza @sergio-teruel review please


pedrobaeza:
Please check CI

payment_sequra/data/payment_provider_data.xml
```diff
@@ -0,0 +1,33 @@
...
+            name="image_128"
+            type="base64"
+            file="payment_sequra/static/description/icon.png"
+        />
+        <field name="code">sequra</field>
+        <field name="company_id" ref="base.main_company" />
+        <field name="redirect_form_view_id" ref="sequra_form" />
+        <field name="sequra_merchant_id">sequra_merchant_id</field>
+        <field name="sequra_account_key">sequra_account_key</field>
+        <field name="sequra_account_secret">sequra_account_secret</field>
```
carlos-lopez-tecnativa, line 24: Why do you fill these fields if they’re invalid? I think it’s better to keep them empty, the user might imagine the payment provider is ready to use.

By the way, could you provide the credentials to test the module?

payment_sequra/data/payment_provider_data.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <record id="payment_method_pp3" model="payment.method">
+        <field name="name">Part payments</field>
+        <field name="code">pp3</field>
```
carlos-lopez-tecnativa, line 5: Any reason to use a code different from the provider code? Normally, both codes can be the same.

payment_sequra/models/payment_provider.py
```diff
@@ -0,0 +1,94 @@
...
+                lambda c: c.name in const.SUPPORTED_CURRENCIES
+            )
+        return supported_currencies
+
+    def _sequra_make_request(self, endpoint, payload=None, method="POST"):
+        self.ensure_one()
+        base_url = (
+            "https://sandbox.sequrapi.com"
+            if self.state == "test"
+            else "https://api.sequrapi.com"
```
carlos-lopez-tecnativa: This base URL could be extracted into a method and reused in other places. I see it repeated in two places

payment_sequra/models/payment_provider.py
```diff
@@ -0,0 +1,94 @@
...
+        )
+        url = endpoint if endpoint.startswith("http") else f"{base_url}{endpoint}"
+        if not (
+            self.sequra_account_key
+            and self.sequra_account_secret
+            and self.sequra_merchant_id
+        ):
+            raise UserError(self.env._("SeQura credentials are not fully configured."))
```
carlos-lopez-tecnativa: Is this validation necessary? The fields are already required at the view level.

payment_sequra/models/payment_transaction.py
```diff
@@ -0,0 +1,257 @@
...
+                    or "127.0.0.1",
+                    "user_agent": environ.get("HTTP_USER_AGENT"),
+                },
+                "delivery_method": {
+                    "name": getattr(
+                        getattr(self.sale_order_ids[:1], "carrier_id", None),
+                        "name",
+                        None,
+                    )
+                    or "Standard delivery",
```
carlos-lopez-tecnativa: This code is hardcoded, and theoretically a payment can be made to a sales order but also to an invoice. So, if I try to pay an invoice, this fails.

payment_sequra/models/payment_transaction.py
```diff
@@ -0,0 +1,257 @@
...
+                "invoice_address": {
+                    "given_names": self.partner_name,
+                    "surnames": self.partner_name,
+                    "company": self.partner_id.company_name or "N/A",
+                    "address_line_1": self.partner_address or "N/A",
+                    "address_line_2": self.partner_id.street2 or "N/A",
+                    "postal_code": self.partner_zip or "00000",
+                    "city": self.partner_city or "Unknown",
+                    "country_code": self.partner_country_id.code or "ES",
+                },
```
carlos-lopez-tecnativa, line 152: If you limit it only to processing sale.order, this model has three fields related to `res.partner`: `partner_id`, `partner_invoice_id`, and `partner_shipping_id`. Could you use these fields in each section?

payment_sequra/models/payment_transaction.py
```diff
@@ -0,0 +1,257 @@
...
+                    response.text,
+                )
+        except Exception:
+            _logger.exception("Error sending PUT to SeQura for tx %s", self.reference)
+        return response
+
+    def _process_notification_data(self, notification_data):
+        super()._process_notification_data(notification_data)
+        if self.provider_code != "sequra":
+            return
```
carlos-lopez-tecnativa, line 229:
```suggestion
        if self.provider_code != "sequra":
            return super()._process_notification_data(notification_data)
```

carlos-lopez-tecnativa (changes requested):
Code review

payment_sequra/data/payment_provider_data.xml
juancarlosonate-tecnativa: I used the same technique as in payment_redsys, fields with `required_if_provider` cause an error during installation if they are empty. I’ll share the credentials with you privately.

payment_sequra/data/payment_provider_data.xml
juancarlosonate-tecnativa: I don’t agree with this. For example, in payment_redsys, a method and a provider are defined with different codes. In the end, a provider can have multiple payment methods, right?

payment_sequra/models/payment_transaction.py
juancarlosonate-tecnativa: This is required by SeQura. If there’s no carrier, it will set "Standard delivery". Should I do it differently?

payment_sequra/models/payment_transaction.py
juancarlosonate-tecnativa: I think this is not correct: https://github.com/odoo/odoo/blob/4fca401148ed798b9c1d04674b44c3287ded5679/addons/payment_paypal/models/payment_transaction.py#L142

payment_sequra/models/payment_transaction.py
juancarlosonate-tecnativa: done

payment_sequra/models/payment_transaction.py
```diff
@@ -0,0 +1,260 @@
...
+            return res
+        payload = self._sequra_prepare_order_request_payload()
+        _logger.info("Creating SeQura order with payload:\n%s", pprint.pformat(payload))
+        response = self.provider_id._sequra_make_request("/orders", payload=payload)
+        location = response.headers.get("Location")
+        if not location:
+            _logger.error(
+                "SeQura did not return a Location header. Response: %s", response.text
+            )
+            return res
```
carlos-lopez-tecnativa: In case of an error, I think you should raise an exception; otherwise, the user won’t know what happened.
In this case, I tested it locally, and in the log I see this error, but on the website, the page just reloads without showing any message.
image

payment_sequra/controllers/main.py
```diff
@@ -0,0 +1,57 @@
...
+        return request.make_response("OK", status=200)
+
+    @http.route(_abort_url, type="http", methods=["GET"], auth="public", csrf=False)
+    def sequra_abort(self, **data):
+        _logger.warning(
+            "SeQura checkout aborted (from %s):\n%s",
+            request.httprequest.remote_addr,
+            pprint.pformat(data),
+        )
+        return request.redirect("/shop/payment")
```
carlos-lopez-tecnativa: In case the user cancels or aborts the transaction, I think it’s better to reflect the same in Odoo, cancel the transaction before returning to this URL.
That way, Odoo will display a proper cancellation message.

image

image

payment_sequra/data/payment_provider_data.xml
carlos-lopez-tecnativa: In most cases, a provider has only one payment method, so the code can be the same. When a provider has more than one payment method, the codes are different. Anyway, it just caught my attention, it’s fine as it is; I just wanted to keep the codes unified or at least related (for example, something like `sequra_pp3`) so it’s clear that they belong to SeQura.

payment_sequra/models/payment_transaction.py
carlos-lopez-tecnativa: It’s the same, just with one line less of code, but anyway.

carlos-lopez-tecnativa (changes requested):

payment_sequra/models/payment_transaction.py
carlos-lopez-tecnativa: I understand this is required by SeQura, but I’d suggest extracting it into a separate helper method (for example, `_get_sequra_delivery_name`()) to make it easier to inherit later if needed.
Also, it would be better to make `Standard delivery` translatable using _() to support localization.

payment_sequra/controllers/main.py
juancarlosonate-tecnativa: done! with dynamic abort url

carlos-lopez-tecnativa (approved):
LGTM

Just a doubt, why is this module in this repository and not in the e-commerce repository?

pedrobaeza:
@carlos-lopez-tecnativa you can generate payment links in the backend and don't use website nor e-commerce at all.

carlos-lopez-tecnativa:
> @carlos-lopez-tecnativa you can generate payment links in the backend and don't use website nor e-commerce at all.

Ok, you’re right. But in that case, please adapt it to work with invoices too, not only with sale orders.
In customer invoices, you can generate a payment link, but it fails.
I mentioned this in https://github.com/OCA/account-payment/pull/901#discussion_r2500494511

image

payment_sequra/models/payment_provider.py
```diff
@@ -0,0 +1,104 @@
...
+    sequra_merchant_id = fields.Char(
+        required_if_provider="sequra", groups="base.group_system"
+    )
+    sequra_account_key = fields.Char(
+        required_if_provider="sequra", groups="base.group_system"
+    )
+    sequra_account_secret = fields.Char(
+        required_if_provider="sequra", groups="base.group_system"
+    )
+    sequra_asset_key = fields.Char(groups="base.group_system")
```
carlos-lopez-tecnativa: This field is unused. What’s the purpose of adding a field that isn’t used?
If it’s meant for another module, then the necessary fields should be added in that other module instead.

payment_sequra/models/payment_transaction.py
```diff
@@ -0,0 +1,290 @@
...
+
+    def _sequra_prepare_order_request_payload(self):
+        # https://docs.sequrapi.com/checkout/order_documentation.html
+        self.ensure_one()
+        base_url = self.provider_id.get_base_url()
+        notify_url = urls.url_join(base_url, SequraController._notify_url)
+        return_url = urls.url_join(base_url, SequraController._return_url)
+        abort_url = urls.url_join(
+            base_url, f"{SequraController._abort_url}/{self.reference}"
+        )
```
carlos-lopez-tecnativa, line 86: I think it’s necessary to `slug/unslug` the `payment.transaction`, or sanitize/replace the reference. When using the invoice sequence, it contains a /, which is interpreted as another endpoint in the URL.
As a result, when I cancel the payment, I get a 404 error.
image

carlos-lopez-tecnativa (changes requested):

payment_sequra/models/payment_transaction.py
pedrobaeza: How is this used in another payment providers? I don't think it's good to use a sequence/reference here, but an absolute identifier that can't be duplicated. For example, `sale.order,25` (like the reference field).

carlos-lopez-tecnativa: > How is this used in another payment providers? I don't think it's good to use a sequence/reference here, but an absolute identifier that can't be duplicated. For example, `sale.order,25` (like the reference field).

I remember, according to @juancarlosonate-tecnativa’s conversation, that SeQura doesn’t send any information when the payment is canceled or aborted.
So, an alternative was to use a URL that includes some data (in this case, the transaction reference) to be able to find the original transaction.

However, this solution can fail if the reference contains a /.

In other modules, you can slug a record, similar to what’s done in the shop module:
https://github.com/odoo/odoo/blob/9e17050696cf5d4b5bdb3a9a37c9b48e5ee63635/addons/website_sale/controllers/main.py#L231

The route could use `payment.transaction` as an absolute identifier, for example:
`"/payment/sequra/abort/<model('payment.transaction'):transaction>"`

And the URL can be built using: `self.env['ir.http']._slug(transaction)`

payment_sequra/models/payment_provider.py
juancarlosonate-tecnativa: Okay, i will remove them and add them in: https://github.com/OCA/e-commerce/pull/1132

payment_sequra/models/payment_transaction.py
juancarlosonate-tecnativa: done, thx!

carlos-lopez-tecnativa (approved):
LGTM

payment_sequra/models/stock_picking.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def _action_done(self):
+        res = super()._action_done()
+        self._sequra_notify_shipment()
+        return res
+
+    def _sequra_notify_shipment(self):
+        # Notify SeQura when the full order is shipped.
+        for picking in self.filtered("sale_id"):
+            transactions = self.env["payment.transaction"].search(
```
victoralmau:
```suggestion
            transactions = self.env["payment.transaction"].sudo().search(
```

victoralmau: It is necessary to use `.sudo()` because when a picking process is completed, the user performing it likely does not have advanced permissions and therefore does not have access to `payment.transaction`.

victoralmau (changes requested):

payment_sequra/models/stock_picking.py
juancarlosonate-tecnativa: Done, thx!

PR #917, opened 2025-12-17
[18.0][FIX] account_payment_return: Missed is_refund key in _compute_payments_widget_reconciled_info...

carlosdauden:
… vals

Seems that the keys used by _compute_payments_widget_reconciled_info were not taken into account during migration.
A comment has been added so that this can be taken into account in future migrations (if anyone reads it...).

TT59860


carlosdauden:
ping @alvaroGC-S @victoralmau

pedrobaeza (approved):

victoralmau (approved):

Issue #919, opened 2026-01-14
Payment Terms calculation issue with Month option 


VBNext:
## Module
payment_terms_extension

## Describe the bug
When a payment term is set to “Months after end of month”, the due date is incorrectly calculated based on the day number of the invoice date’s month, instead of the actual end of the target month.

## To Reproduce

**Affected versions**:

- [ ] 17.0
- [ ] 18.0: https://github.com/OCA/account-payment/pull/920
- [ ] 19.0 ?

1.  Configure a payment term with 1 Month after end of month.

<img width="1134" height="272" alt="Image" src="https://github.com/user-attachments/assets/9acaa94f-7220-4ab8-beb7-18f400dcc3ac" />

2. Set Example date with a date in a month that has fewer days (e.g. February with 28 days).

**Expected behavior**
The due date should be set to the end of the following month, for example 03/31/2025, and not 03/28/2025

PR #722, opened 2024-03-26
[17.0] [MIG] account_payment_term_extension

AnaisLForgeFlow:
Standard migration

@ForgeFlow

PR #920, opened 2026-01-14
[18.0] [FIX] account_payment_term_extension: put end of month calculation last

odooNextev:
https://github.com/OCA/account-payment/issues/919

### DIFF
--- a/account_payment_term_extension/models/account_payment_term_line.py
+++ b/account_payment_term_extension/models/account_payment_term_line.py
@@ -117,8 +117,8 @@ class AccountPaymentTermLine(models.Model):
         elif self.delay_type == "months_after":
             return due_date + relativedelta(months=self.nb_days)
         elif self.delay_type == "months_after_end_of_month":
-            return date_utils.end_of(due_date, "month") + relativedelta(
-                months=self.nb_days
+            return date_utils.end_of(
+                due_date + relativedelta(months=self.nb_days), "month"
             )
         return res
 
--- a/account_payment_term_extension/tests/test_payment_term.py
+++ b/account_payment_term_extension/tests/test_payment_term.py
@@ -46,41 +46,38 @@ class TestAccountPaymentTerm(BaseCommon):
         )
 
     def test_01_compute_terms(self):
-        delay_setting = self.env.company.payment_terms_delay_type
-        payment_terms_delay_type = delay_setting or False
-        if payment_terms_delay_type:
-            if payment_terms_delay_type == "weeks":
-                two_week_payterm = self.account_payment_term.create(
-                    {
-                        "name": "2 weeks",
-                        "line_ids": [
-                            fields.Command.create(
-                                {
-                                    "value": "percent",
-                                    "value_amount": 100.0,
-                                    "nb_days": 2,
-                                    "delay_type": "weeks_after",
-                                },
-                            ),
-                        ],
-                    }
-                )
-                res = two_week_payterm._compute_terms(
-                    currency=self.currency,
-                    company=self.company,
-                    date_ref="2015-03-02",
-                    tax_amount=0,
-                    tax_amount_currency=0,
-                    sign=1,
-                    untaxed_amount=100,
-                    untaxed_amount_currency=100,
-                )
-                payment_term_date = self._get_date_payment_term_line(res["line_ids"][0])
-                self.assertEqual(
-                    payment_term_date,
-                    "2015-03-16",
-                    "Error in the _compute_terms of payment terms with weeks",
-                )
+        self.env.company.payment_terms_delay_type = "weeks"
+        two_week_payterm = self.account_payment_term.create(
+            {
+                "name": "2 weeks",
+                "line_ids": [
+                    fields.Command.create(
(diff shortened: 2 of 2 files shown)


VBNext (approved):
Functional OK

odooNextev:
Hi @RicardCForgeFlow,
We noticed you migrate `account_payment_term_extension` to 17.0 with this commit and this added 
```
elif self.delay_type == "months_after_end_of_month":
            return date_utils.end_of(due_date, "month") + relativedelta(
                months=self.nb_days
```
https://github.com/OCA/account-payment/commit/1c0631740c6f04a11d3b198b3cd486d53462330d#diff-03c97f40a26930117d566a7070a68b0c747ce8f00b3f3cabd051adcb61a1c1bfR118

We guess `end_of` should be calculated last, otherwise the end date day won't always be the end of the month following the given months, but will depend on the last day of the starting month, but the days in the months aren't always the same.

We opened an issue https://github.com/OCA/account-payment/issues/919 and I did this PR for 18.0 that carry on this bug.

I'd ask you if you'd approve this change.

@TheMule71 can you check this little fix too?
We saw you migrate this module to 18.0.

RicardCForgeFlow:
> Hi @RicardCForgeFlow, We noticed you migrate `account_payment_term_extension` to 17.0 with this commit and this added
> 
> ```
> elif self.delay_type == "months_after_end_of_month":
>             return date_utils.end_of(due_date, "month") + relativedelta(
>                 months=self.nb_days
> ```
> 
> [1c06317#diff-03c97f40a26930117d566a7070a68b0c747ce8f00b3f3cabd051adcb61a1c1bfR118](https://github.com/OCA/account-payment/commit/1c0631740c6f04a11d3b198b3cd486d53462330d#diff-03c97f40a26930117d566a7070a68b0c747ce8f00b3f3cabd051adcb61a1c1bfR118)
> 
> We guess `end_of` should be calculated last, otherwise the end date day won't always be the end of the month following the given months, but will depend on the last day of the starting month, but the days in the months aren't always the same.
> 
> We opened an issue #919 and I did this PR for 18.0 that carry on this bug.
> 
> I'd ask you if you'd approve this change.
> 
> @TheMule71 can you check this little fix too? We saw you migrate this module to 18.0.

Hello @odooNextev I agree. 

Could you modify the: https://github.com/OCA/account-payment/blob/b220020f97353cd11e62744d531ba9e219553d6d/account_payment_term_extension/tests/test_payment_term.py#L105
to test that case? For example:
`date_ref="2015-02-02"`
and the payment_term_date should be `2015-04-31`

Also, I think the tests should be modified to change the `payment_terms_delay_type` of the company in every test, right?

Thank you!

I have to say that I don't know how am I the author of the migration commit, if I did nothing in that PR: https://github.com/OCA/account-payment/pull/722

:laughing:

odooNextev:
> Hello @odooNextev I agree.
> 
> Could you modify the:
> 
> https://github.com/OCA/account-payment/blob/b220020f97353cd11e62744d531ba9e219553d6d/account_payment_term_extension/tests/test_payment_term.py#L105
> 
> 
> to test that case? For example:
> `date_ref="2015-02-02"`
> and the payment_term_date should be `2015-04-31`

Hi Ricard,
thank you for your quick reply.
I changed the date as you suggested me to make sure the problem doesn't appear again.

> Also, I think the tests should be modified to change the `payment_terms_delay_type` of the company in every test, right?
> 
> Thank you!

You're right, I fixed this and added some tests to check that the `delay_type` selection respect the `payment_terms_delay_type` setting.

RicardCForgeFlow (approved):

RicardCForgeFlow:
Code Review 👍

odooNextev:
> Code Review 👍

Thank you! 🙏

odooNextev:
@OCA/banking-maintainers what do you think about this fix?
This is important cause without it, the end of the month can be calculated incorrectly and tests aren't very effective.
We also created an issue because it needs to be ported to 17.0 and 19.0.

odooNextev:
Hi @pedrobaeza,
what do you think about this fix?
This is important cause without it, the end of the month can be calculated incorrectly and tests aren't very effective.
We also created an issue because it needs to be ported to 17.0 and 19.0.

pedrobaeza:
Changing existing tests doesn't sound good, as that means that the expected behavior is changing. Can you explain that change?

odooNextev:
> Changing existing tests doesn't sound good, as that means that the expected behavior is changing. Can you explain that change?

As suggested me @RicardCForgeFlow and as I noticed debugging the tests, many tests get `delay` from `self.env.company.payment_terms_delay_type`, but this is never set and the condition in these lines (for example) will always be false: https://github.com/OCA/account-payment/blob/18.0/account_payment_term_extension/tests/test_payment_term.py#L49-L51

Unless I'm missing something, the tests I modified never worked.

stenext (approved):
LGTM

StefanRijnhart (approved):
/ocabot merge patch

stenext:
> /ocabot merge patch

@StefanRijnhart Thanks

PR #923, opened 2026-01-27
[18.0][ADD] account_payment_internal_transfer

JoanSForgeFlow:
This module restores the internal transfer functionality for payments that was available in previous Odoo versions but was removed.

Now the recommended approach for internal transfers requires waiting for bank statement lines to be imported an then using reconciliation models. This workflow has significant drawbacks:

- Users cannot register transfers at the moment they occur
- Users without direct bank access have no way to track pending transfers

This module allows users to register internal transfers immediately when they happen, creating the intermediate journal entries with outstanding accounts. Later, when bank statements are imported, each side can be reconciled
independently. This results in 4 journal entries total (2 payments + 2 bank statement reconciliations).


JoanSForgeFlow:
Hi @AaronHForgeFlow,

Do you know if we can reopen this PR?

It has been migrated also to 19.0 and I think is very useful https://github.com/OCA/account-payment/pull/959

Thank you

PR #926, opened 2026-02-24
[18.0][FIX] account_payment_line: Use vals.get() instead of direct dict access

Saran440:
Test Script error when install `account_payment_line` module

### DIFF
--- a/account_payment_line/models/account_payment.py
+++ b/account_payment_line/models/account_payment.py
@@ -121,7 +121,7 @@ class AccountPayment(models.Model):
         filter_vals_list = [
             vals
             for vals in vals_list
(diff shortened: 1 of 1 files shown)


celm1990:
The error comes from this commit in Odoo: https://github.com/odoo/odoo/commit/531b887aec92c2fbf57495992be9fbc32d9ea20e
So, this change only hides the error, but the correct fix is to use `balance` instead of `credit` and `debit`.

celm1990:
Please see the fix in https://github.com/OCA/account-payment/pull/927

Saran440:
move to https://github.com/OCA/account-payment/pull/927

PR #930, opened 2026-04-16
[18.0][FIX] account_due_list: deprecated invisible attr in payment_view

cristina-hidalgo-tecnativa:
Removed the `parent_state` field declaration from `view_payments_tree`, as Odoo 18 auto-populates fields used in `decoration-*` expressions without needing an explicit declaration.

@Tecnativa TT62027
@pedrobaeza @carlosdauden

### DIFF
--- a/account_due_list/views/payment_view.xml
+++ b/account_due_list/views/payment_view.xml
@@ -64,7 +64,6 @@
                 <field name="date_maturity" optional="show" />
                 <field name="reconciled" readonly="1" optional="hide" />
                 <field name="full_reconcile_id" readonly="1" optional="hide" />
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
In fact, you can remove it at all, as now Odoo auto-populates fields that are used in other parts of the view.

cristina-hidalgo-tecnativa:
> In fact, you can remove it at all, as now Odoo auto-populates fields that are used in other parts of the view.

Removed field.

CI test failures are unrelated to this change — It seems they belong to `account_payment_line` test [#931 ](https://github.com/OCA/account-payment/pull/931)

carlosdauden (approved):

PR #931, opened 2026-04-16
[18.0][FIX] account_payment_line: update payment state assertions

cristina-hidalgo-tecnativa:
Update `test_03` and `test_04` assertions from `assertEqual(state, "in_process")` to `assertIn(state, ["in_process", "paid"])`, as in 18 version the resulting state depends on the journal configuration — whether an outstanding payments account is used or not.

@Tecnativa TT62027
@pedrobaeza @christian-ramos-tecnativa


pedrobaeza (approved):
Let's solve the problem in the branch.

/ocabot merge patch

I'm not sure though that this module should do that asserts if it doesn't touch this part.

PR #932, opened 2026-04-19
[18.0][FIX] account_payment_term_extension: fix percentage calculation with fixed amount on first line

### LINKED ISSUE #924: [18.0] account_payment_term_extension - Fixed amount on first line breaks percentage calculation
**Module**

account_payment_term_extension

**Describe the bug**

When a fixed amount is set on the first line of a payment term, the amount is not subtracted from the invoice total before calculating the percentage-based lines.

As a result, the percentage lines are calculated on the full invoice total instead of the residual amount (total minus fixed amount), which leads to an incorrect final installment. In some cases, this produces a negative amount on the last payment term line.

**To Reproduce**

Affected versions:

Odoo 18.0

Module: account_payment_term_extension

Create a payment term with:

```
Line 1: Fixed amount = 400€, Due immediately (0 days)
Line 2: 33% at 30 days
Line 3: 33% at 60 days
Line 4: 34% at 90 days

```

Create a customer invoice for 1,000€.

Apply the payment term to the invoice.

Expected behavior

The fixed amount (400€) should be subtracted from the invoice total before applying the percentage calculations.

Correct expected calculation for a 1,000€ invoice:

```
400€ at day 0
198€ at 30 days (33% of 600€)
198€ at 60 days (33% of 600€)
204€ at 90 days (34% of 600€)
```

Total: 1,000€

Actual behavior

The module calculates the percentage lines on the full invoice amount (1,000€), resulting in:

```
400.00€ due immediately
330.00€ at 30 days
330.00€ at 60 days
-60.00€ at 90 days
```

The last installment becomes negative because the total of previous lines exceeds the invoice total.

fsmw:
## Description

When a payment term has a fixed amount on the first line and percentage lines after it, the percentage lines are calculated on the full invoice total instead of the remaining amount after deducting the fixed amount.

This causes incorrect amounts and in some cases negative amounts on the last payment term line.

## Steps to reproduce

1. Create a payment term with:
   - Line 1: Fixed amount = 400€, Due immediately (0 days)
   - Line 2: 33% at 30 days
   - Line 3: 33% at 60 days
   - Line 4: 34% at 90 days

2. Create an invoice for 1000€ with this payment term

**Expected:**
- 400€ at day 0 (fixed)
- 198€ at 30 days (33% of 600€)
- 198€ at 60 days (33% of 600€)
- 204€ at 90 days (34% of 600€)

**Actual (bug):**
- 400€ at day 0
- 330€ at 30 days (33% of 1000€)
- 330€ at 60 days (33% of 1000€)
- -60€ at 90 days (negative!)

## Fix

Changed  in both  and  to use  instead of  for percentage calculations. This ensures that percentage lines are calculated on the residual amount after fixed amounts have been deducted.

## Related Issue

Closes https://github.com/OCA/account-payment/issues/924

---

Contributor from SMWLAB community. First OCA PR!

### DIFF
--- a/account_payment_term_extension/models/account_payment_term.py
+++ b/account_payment_term_extension/models/account_payment_term.py
@@ -183,12 +183,13 @@ class AccountPaymentTerm(models.Model):
                 term_vals["company_amount"] = company_line_amount
                 term_vals["foreign_amount"] = line_amount
             else:
-                # Percentage amounts
+                # Percentage amounts - use remaining_amount for correct calculation
+                # when there are fixed amounts on previous lines
                 line_amount = line.compute_line_amount(
-                    total_amount, remaining_amount, precision_digits
+                    remaining_amount, remaining_amount, precision_digits
                 )
                 company_line_amount = line.compute_line_amount(
-                    total_amount_currency,
+                    remaining_amount_currency,
                     remaining_amount_currency,
                     company_precision_digits,
                 )
--- a/account_payment_term_extension/models/account_payment_term_line.py
+++ b/account_payment_term_extension/models/account_payment_term_line.py
@@ -150,7 +150,8 @@ class AccountPaymentTermLine(models.Model):
         if self.value == "fixed":
             return float_round(self.value_amount, precision_digits=precision_digits)
         elif self.value in ("percent", "percent_amount_untaxed"):
-            amt = total_amount * self.value_amount / 100.0
+            # Use remaining_amount to correctly handle fixed amounts on previous lines
+            amt = remaining_amount * self.value_amount / 100.0
             if self.amount_round:
                 amt = float_round(amt, precision_rounding=self.amount_round)
             return float_round(amt, precision_digits=precision_digits)


fsmw:
Hi @pedrobaeza — would you have a moment to review this fix?

When a payment term has a fixed amount on the first line and percentage lines after, the percentages were being calculated on the full total instead of the remaining amount after deducting the fixed amount. This caused incorrect and sometimes negative amounts.

The fix changes `total_amount` → `remaining_amount` for percentage calculations (2 files, 10 lines). Includes the existing test suite coverage. CI is all green.

This has been open since April — any feedback would be appreciated. Thanks!

pedrobaeza (commented):
Please add a test exercising the problem

JuanJaraSandoval:
@fsmw @pedrobaeza — heads-up on an overlap.

#949 reports a different bug in the same block: in multi-currency,
`company_amount` is computed from `total_amount_currency` and `foreign_amount`
from `total_amount`, with the two rounding precisions crossed as well. Since
`residual_amount` is then decremented with a document-currency magnitude, the
balance line ends up with `company_amount` and `foreign_amount` of opposite
signs, and posting fails on
`account_move_line_check_amount_currency_balance_sign`.

The two fixes are independent, so I would rather not add scope here. I am happy
to take the multi-currency one in a separate PR with its regression test, and to
rebase on whichever lands first — just flagging it so we don't touch the same
lines twice without knowing.

On this PR, in case it helps unblock it: with both hunks, percentages end up
being computed from `remaining_amount`. But `remaining_amount` is assigned once,
at `total_amount = remaining_amount = tax_amount + untaxed_amount`, and is never
decremented — only `residual_amount` is, at the end of the loop. So it still
equals `total_amount` and the computed value is unchanged.

The docstring of `compute_line_amount` already describes the intended behaviour:

    :param remaining_amount: total amount minus sum of previous lines
        computed amount

so that parameter was always meant to be the running remainder; what is missing
is the caller maintaining it. Decrementing `remaining_amount` inside the loop —
or using `residual_amount`, which is already maintained — would make the fix
effective, and would also avoid changing the meaning of the first parameter of a
public method.

If it helps, I can post a draft of the regression test for the scenario in #924
as a starting point.

pedrobaeza:
OK, go ahead

JuanJaraSandoval:
Opened #980 for the multi-currency fix (#949), with a regression test. It touches the same block in _compute_terms, so whichever lands second will need a rebase — happy to do it on my side.

PR #934, opened 2026-04-20
[18.0][ADD] payment_amazon_pay: Add new module for Amazon Pay

pilarvargas-tecnativa:
@Tecnativa TT61807


juancarlosonate-tecnativa:
ping @carlos-lopez-tecnativa

payment_amazon_pay/controllers/main.py
```diff
@@ -0,0 +1,209 @@
...
+    @staticmethod
+    def _find_tx(reference=None, checkout_session_id=None, charge_id=None):
+        domain = [("provider_code", "=", "amazon_pay")]
+        if reference:
+            domain.append(("reference", "=", reference))
+        elif checkout_session_id:
+            domain.append(("amazon_pay_checkout_session_id", "=", checkout_session_id))
+        elif charge_id:
+            domain.append(("amazon_pay_charge_id", "=", charge_id))
+        return request.env["payment.transaction"].sudo().search(domain, limit=1)
```
carlos-lopez-tecnativa: I’m not sure if this can be unified with the `_get_tx_from_notification_data` function. I see some code is partially duplicated.

payment_amazon_pay/controllers/main.py
```diff
@@ -0,0 +1,209 @@
...
+      if (typeof amazon !== 'undefined') {{
+        render();
+      }} else {{
+        document.querySelector('script[src="{checkout_js_url}"]')
+          .addEventListener('load', render);
+      }}
+    }})();
+  </script>
+</body>
+</html>"""
```
carlos-lopez-tecnativa: Why is all the code embedded here? I imagine this can be done in a QWeb view and JavaScript files instead.

payment_amazon_pay/controllers/main.py
```diff
@@ -0,0 +1,209 @@
...
+    .box {{
+      background:white;padding:40px;border-radius:8px;
+      box-shadow:0 2px 8px rgba(0,0,0,.1);text-align:center;max-width:400px;
+    }}
+    p {{color:#555;margin-bottom:24px;}}
+  </style>
+</head>
+<body>
+  <div class="box">
+    <p>Haz clic en el botón para completar el pago con Amazon Pay.</p>
```
carlos-lopez-tecnativa: This text is not translatable in this file; if you move it to QWeb, I assume it can be translated.

carlos-lopez-tecnativa (changes requested):
Code review

payment_amazon_pay/controllers/main.py
cristina-hidalgo-tecnativa: Removed the duplicate code

payment_amazon_pay/controllers/main.py
cristina-hidalgo-tecnativa: Done.

payment_amazon_pay/controllers/main.py
cristina-hidalgo-tecnativa: Fixed. The text is now in the QWeb template and appears in the .pot file. A Spanish translation (`i18n/es.po`) has also been added.

payment_amazon_pay/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "Amazon Pay",
+    "version": "18.0.1.0.0",
+    "category": "Accounting/Payment Providers",
+    "summary": "Amazon Pay payment provider for Odoo 18",
```
carlos-lopez-tecnativa: I think it is not recommended to include the Odoo version here, especially if this module is migrated to v19+ in the future.
```suggestion
    "summary": "Amazon Pay payment provider for Odoo",
```

payment_amazon_pay/models/payment_provider.py
```diff
@@ -0,0 +1,398 @@
...
+                data = "\n".join(data[i : i + 64] for i in range(0, len(data), 64))
+                key = f"{header}\n{data}\n{footer}"
+        return key
+
+    def _amazon_pay_build_headers(
+        self, method, api_path, payload_text="", idempotency_key=None
+    ):
+        self.ensure_one()
+        if not self.amazon_pay_private_key:
+            raise ValidationError(_("Amazon Pay private key is missing."))
```
carlos-lopez-tecnativa: Check all occurrences, please.

```suggestion
            raise ValidationError(self.env._("Amazon Pay private key is missing."))
```

payment_amazon_pay/models/payment_transaction.py
```diff
@@ -0,0 +1,132 @@
...
+    def _process_notification_data(self, notification_data):
+        super()._process_notification_data(notification_data)
+        if self.provider_code != "amazon_pay":
+            return
+
+        payment_method = self.env["payment.method"].search(
+            [("code", "=", "amazon_pay")], limit=1
+        )
+        if payment_method:
+            self.payment_method_id = payment_method
```
carlos-lopez-tecnativa: Why do you need to fill the `payment_method_id` here?
The Odoo base should populate this field in https://github.com/odoo/odoo/blob/678f2a48f66ef0bfbf4fdbbaf65fdf11fd499f05/addons/payment/controllers/portal.py#L358.
This value should come from the option selected by the user. Imagine if the payment provider has more than one payment method (this may not be the case here), but theoretically, a payment provider can have more than one payment method.

payment_amazon_pay/controllers/main.py
```diff
@@ -0,0 +1,173 @@
...
+            )
+            return request.redirect("/payment/status")
+
+        try:
+            btn = tx.provider_id._amazon_pay_get_button_payload(tx)
+        except Exception:
+            _logger.exception(
+                "Amazon Pay: failed to build button payload for %s", reference
+            )
+            return request.redirect("/payment/status")
```
carlos-lopez-tecnativa: I'm not sure if simply redirecting to /payment/status is OK. If I have an error, it is not displayed to the user, and they may think everything is fine when it is not.

In my case, I don't have the credentials to test this module functionally, but I entered dummy credentials, and when I pay on the website, no error is displayed; it is just redirected to /payment/status, and the user can end up waiting there without any message describing what really happened.

Image

carlos-lopez-tecnativa (changes requested):
Just a minor comment on the technical side: please review my comment about displaying errors if something occurs. I imagine that if the credentials are incorrect, the API should return some errors which can be displayed to the user. What do you think about the user experience?

payment_amazon_pay/__manifest__.py
cristina-hidalgo-tecnativa: Fixed

payment_amazon_pay/models/payment_provider.py
cristina-hidalgo-tecnativa: Done, updated all occurrences in both model files.

payment_amazon_pay/models/payment_transaction.py
cristina-hidalgo-tecnativa: You're right, removed. The base already sets it when creating the transaction.

payment_amazon_pay/controllers/main.py
cristina-hidalgo-tecnativa: Fixed. When building the button payload fails, `_set_error` is now called on the transaction before redirecting, so the status page reflects the actual error state.

cristina-hidalgo-tecnativa:
> Just a minor comment on the technical side: please review my comment about displaying errors if something occurs. I imagine that if the credentials are incorrect, the API should return some errors which can be displayed to the user. What do you think about the user experience?

Good point. I've updated the error handling so that API errors are now displayed to the user on the status page instead of being silently swallowed.

payment_amazon_pay/controllers/main.py
```diff
@@ -0,0 +1,178 @@
...
+            btn = tx.provider_id._amazon_pay_get_button_payload(tx)
+        except ValidationError as e:
+            _logger.warning("Amazon Pay: button payload error for %s: %s", reference, e)
+            tx._set_error(str(e))
+            return request.redirect(tx.landing_route or "/payment/status")
+        except Exception:
+            _logger.exception(
+                "Amazon Pay: failed to build button payload for %s", reference
+            )
+            tx._set_error("Amazon Pay: failed to prepare the payment button.")
```
carlos-lopez-tecnativa: I’m not sure if you intentionally didn’t add an exception message here to avoid displaying technical information to the user, but since you do it in the `ValidationError`, it could be added here as well. Anyway, just a comment.


```suggestion
        except Exception as ex:
            _logger.exception(
                "Amazon Pay: failed to build button payload for %s", reference
            )
            tx._set_error(f"Amazon Pay: failed to prepare the payment button. {str(ex)}")
```

carlos-lopez-tecnativa (approved):
Code review LGTM

payment_amazon_pay/controllers/main.py
cristina-hidalgo-tecnativa: Done. The exception message is now included in both the `ValidationError` and generic `Exception` handlers.

payment_amazon_pay/data/payment_provider_data.xml
```diff
@@ -0,0 +1,31 @@
...
+            <field
+                name="image"
+                type="base64"
+                file="payment_amazon_pay/static/img/amazon_pay.png"
+            />
+            <field name="support_tokenization">False</field>
+            <field name="support_express_checkout">False</field>
+            <field name="support_refund">partial</field>
+        </record>
+    </data>
```
pilarvargas-tecnativa:
```suggestion
    
```
If this module depends on payment, why not use the method defined for Amazon instead of creating a new one?

https://github.com/odoo/odoo/blob/91e660ec8805b40c7d59fa31ba63e2f62900b96c/addons/payment/data/payment_method_data.xml#L267

payment_amazon_pay/data/payment_provider_data.xml
```diff
@@ -0,0 +1,31 @@
...
+        <field name="code">amazon_pay</field>
+        <field
+            name="image_128"
+            type="base64"
+            file="payment_amazon_pay/static/img/amazon_pay.png"
+        />
+        <field name="module_id" ref="base.module_payment_amazon_pay" />
+        <field
+            name="payment_method_ids"
+            eval="[Command.set([ref('payment_amazon_pay.payment_method_amazon_pay')])]"
```
pilarvargas-tecnativa:
```suggestion
            eval="[Command.set([ref('payment.payment_method_amazon_pay')])]"
```

payment_amazon_pay/views/payment_amazon_pay_templates.xml
```diff
@@ -0,0 +1,36 @@
...
+                <script id="o_amazon_pay_sdk" t-att-src="checkout_js_url" />
+                <script type="application/json" id="amazon-pay-cfg">
+                    <t t-out="amazon_pay_cfg" />
+                </script>
+                <script
+                    src="/payment_amazon_pay/static/src/js/amazon_pay_checkout.js"
+                />
+            </body>
+        </html>
+    </template>
```
pilarvargas-tecnativa: This is really necessary

payment_amazon_pay/data/payment_provider_data.xml
```diff
@@ -0,0 +1,16 @@
...
+        <field
+            name="image_128"
+            type="base64"
+            file="payment_amazon_pay/static/img/amazon_pay.png"
+        />
+        <field
+            name="payment_method_ids"
+            eval="[Command.set([ref('payment.payment_method_amazon_pay')])]"
+        />
+    </record>
```
pilarvargas-tecnativa, line 16:
```suggestion
        <field name="module_id" ref="base.module_payment_amazon_pay" />
        <field
            name="payment_method_ids"
            eval="[Command.set([ref('payment.payment_method_amazon_pay')])]"
        />
    </record>
```
The module is important for publishing or unpublishing the provider, since visibility depends on the status of the module provided by that provider
https://github.com/odoo/odoo/blob/a439bd305112f6efc752ce900f7782e7faaf7312/addons/payment/views/payment_provider_views.xml#L14

cristina-hidalgo-tecnativa: Added.

juancarlosonate-tecnativa (changes requested):
Button flow needs review, it shouldn't preload in start() or mutate the shared paymentContext. It should be set up only when the option is selected instead of globally. See payment_paypal for the pattern:

https://github.com/odoo/odoo/blob/18.0/addons/payment_paypal/static/src/js/payment_form.js

pilarvargas-tecnativa:
> Button flow needs review, it shouldn't preload in start() or mutate the shared paymentContext. It should be set up only when the option is selected instead of globally. See payment_paypal for the pattern:

This should have been solved already

PR #935, opened 2026-04-22
[17.0][ADD] account_move_reconcile_export: Make reconciled moves exportable

christian-ramos-tecnativa:
Adds a new computed field `reconciled_move_ids` on `account.move` that exposes all journal entries reconciled against a posted invoice or receipt.

When exporting or reporting invoice data, there is no direct ORM-accessible field to retrieve the journal entries (payments, bank statement lines, etc.) that have been reconciled against a given invoice. This module fills that gap without duplicating the reconciliation logic already present in Odoo core.

@Tecnativa TT62114
@pedrobaeza


pedrobaeza (approved):

Yoana-RB (approved):
Functional review LGTM!

PR #944, opened 2026-05-25
[17.0][FIX] account_payment_promissory_note: Check active_model to compute …

christian-ramos-tecnativa:
…date_due


pedrobaeza (approved):
Let's fast track it to fix it the problem: 

/ocabot merge patch

PR #948, opened 2026-05-26
[19.0][MIG] account_payment_notification

Shide:
Things done:
- Moved `res_config_settings` files to `models/`
- mobile field now is related to `partner_id.phone` instead of `partner_id.mobile`
- translations from `_(` to `self.env._(`
- Fixed tests to check sms bodies with `Markup.striptags()`

payment_notif

MT-14666 @moduon @yajo @EmilioPascual @chienandalu @fcvalgar please review if you want 😄


chienandalu (approved):
Code review 👍

fcvalgar (changes requested):
Thank you for the work on this contribution @Shide . I have tested the module and found one point that I would like to clarify before being able to approve it.

The automatic notification flow seems to be triggered only when using the Mark as Sent action. However, when a payment is marked as paid, the payment is not automatically marked as sent, so the automatic notification is not triggered in that flow.
Could you please confirm if this is the expected behavior? If the module is only intended to notify when the user explicitly clicks Mark as Sent, then the behavior is consistent with the current implementation. If payments marked as paid should also be considered sent, then this flow seems to be missing.

Tests performed:
Test 1: OK - Manual email notification from the payment form. The email wizard opened correctly with the payment notification template.
Test 2: OK - Manual SMS notification from the payment form. The SMS wizard opened correctly with the payment notification template.
Test 3: OK - Manual partner notification from the payments list. The configured notification method was correctly applied to the selected payment.
Test 4: OK - Automatic notification using Mark as Sent. The notification flow is correctly triggered from the Mark as Sent action.
Test 5: Not OK - Payment marked as paid does not become sent automatically. As a result, the automatic notification is not triggered when the payment is only marked as paid.
Test 6: OK - Required notification with a partner without valid contact data. The payment was correctly blocked when notifications were required and the partner could not be notified.
Test 7: OK - Required notification disabled. The payment could be marked as sent even when the partner could not be notified.

Could you please confirm the expected behavior for payments that are marked as paid but not explicitly marked as sent?

Image

account_payment_notification/readme/CONFIGURE.md
```diff
@@ -0,0 +1,13 @@
...
+
+1.  Go to *Settings*.
+2.  Go to *Invoicing*.
+3.  Choose if you want to *Require payment notifications*.
+4.  Select the desired *Payment notification method*.
+5.  Select the desired *Payment notifications automatism*.
+
+> [!NOTE]
+> Automated notifications will be sent or queued when the payment is
+> marked as *sent*, not as *paid*. A *sent* payment is one that has been sent to
```
Shide, line 11: @fcvalgar the flow of the module is the same that v18.
This part of README says that the payment must be marked as sent.
If you have intermediate accounts that needs to be reconciled with the bank (not the bank directly), the sent state will be triggered.

fcvalgar (approved):
LGTM @Shide 

As a minor documentation suggestion, I think it would be useful to clarify in the README that this module is intended for payment flows where payments are explicitly sent to the bank, such as SEPA Credit Transfer or SEPA Direct Debit flows. Automatic notifications are triggered when the payment is marked as sent, not when it is simply marked as paid, so making this scope explicit would help functional users and reviewers understand the expected behavior more clearly.

BhaveshHeliconia (approved):
Code and functional review LGTM!

rafaelbn (approved):
Functional tested 👍🏼 

/ocabot merge patch

PR #883, opened 2025-09-24
[18.0][FIX] account_payment_return_import_iso20022: payment return match by payment id

qgroulard:
Draft PR to open a discussion.
From my understanding and functional tests, the `payment_id` and not the `move_id` is passed as the End2end_identification from the generation of the payment order to the reception of the return.

I don't understand how can we expect to receive the `move_id` as the line `reference`. Is this a mistake or can someone explain this mechanism to me ?

PR #955, opened 2026-06-23
[18.0][FIX] account_payment_return_import_iso20022: Search payments by ID instead of move_id

carlos-lopez-tecnativa:
After this PR https://github.com/OCA/bank-payment/pull/1475, the `EndToEndId` is generated using the payment ID instead of the journal entry (move) ID.

As a result, when importing an payment return XML file, the payment can no longer be matched correctly because the lookup is still performed using the move_id.

TT63378
@Tecnativa @pedrobaeza @carlosdauden @sergio-teruel could you please review this?

### DIFF
--- a/account_payment_return_import_iso20022/models/payment_return.py
+++ b/account_payment_return_import_iso20022/models/payment_return.py
@@ -16,10 +16,17 @@ class PaymentReturnLine(models.Model):
                 continue
             payments = self.env["account.payment"].search(
                 [
-                    ("move_id", "=", int(line.reference)),
+                    ("id", "=", int(line.reference)),
                     ("payment_order_id", "!=", False),
                 ],
             )
+            if not payments:
+                payments = self.env["account.payment"].search(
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please look for both, as we may have old payment orders.

carlos-lopez-tecnativa:
> Please look for both, as we may have old payment orders.

Done. Please let me know if this is OK for you.

pedrobaeza (commented):
You can better swap the search order. I'm now fearing that there's a choke between the IDs of both tables in any moment (but that's not your fault, but the change of criteria).

pedrobaeza (commented):
Now thinking twice, this may return incorrect results due to the choke I have mentioned. Do we simply break compatibility with older payments? What others think?

pedrobaeza:
Already covered in #883, and finally I think it's better to not search by the old one, as this can provoke false matches. Closing.

PR #956, opened 2026-06-24
[19.0] [MIG] account_due_list_payment: Migration to 19.0

Borruso:
https://github.com/OCA/account-payment/issues/885


account_due_list_payment/tests/test_action_register_payment.py
```diff
@@ -4,7 +4,7 @@
 from odoo.tests import common
 
 
-class TestAccountMoveLine(common.SavepointCase):
+class TestAccountMoveLine(common.TransactionCase):
```
kalpeshgajera-uncannycs: TransactionCase enables mail tracking, which may introduce unnecessary overhead in this test. Please consider inheriting from BaseCommon instead. If TransactionCase is required, please explicitly disable tracking by adding cls.tracking = False in setUpClass.

PR #962, opened 2026-07-27
[18.0][IMP] account_payment_line: restrict counterpart lines to payment partner

rjaraspearhead:
## Context

The counterpart lines of a payment currently accept any partner. This makes it
possible to register a payment for one partner and then apply, in its
counterpart lines, invoices belonging to a completely different partner. Some
companies rely on that to move balances between partners, but for others it is
a source of reconciliation errors that are hard to spot afterwards.

## What this PR does

Adds an optional company setting to forbid it, **disabled by default**, so the
current behaviour is unchanged for everybody who does not opt in:

- New field `res.company.payment_line_restrict_counterpart_partner`, exposed in
  *Invoicing / Settings* through `res.config.settings`.
- A computed domain on the counterpart line restricts the partner selection to
  the commercial entity of the payment partner. When the payment has no partner
  yet, the selection is blocked rather than falling back to every partner.
- An `@api.constrains` on the counterpart line rejects lines whose partner,
  move or journal item belongs to a different commercial entity, so the
  restriction is enforced on save and not only in the UI.

Using the commercial entity means invoices of child contacts of the payment
partner are still accepted, which is the usual expectation.

## How to test

1. Leave the setting off and check that a payment can still apply an invoice of
   another partner (existing behaviour).
2. Enable *Restrict counterpart lines to payment partner* in Invoicing /
   Settings.
3. On a payment for partner A, the partner of the counterpart lines can only be
   A or one of its contacts, and saving a line for partner B raises a
   `ValidationError`.

Covered by `test_12_restrict_counterpart_partner`.


rjaraspearhead:
Hi @celm1990, @ChrisOForgeFlow, @xaviedoanhduy, @AlexPForgeFlow — could you review this
when you have a moment?

account_payment_line/models/account_payment.py
```diff
@@ -20,6 +20,16 @@ class AccountPayment(models.Model):
...
+    restrict_counterpart_partner = fields.Boolean(
+        compute="_compute_restrict_counterpart_partner",
+    )
+
+    @api.depends("company_id.payment_line_restrict_counterpart_partner")
+    def _compute_restrict_counterpart_partner(self):
+        for rec in self:
+            rec.restrict_counterpart_partner = (
+                rec.company_id.payment_line_restrict_counterpart_partner
+            )
```
celm1990: Why do you need a computed field? Wouldn't a simple related field work?

celm1990 (commented):
LGTM, just a technical question.

account_payment_line/models/account_payment.py
rjaraspearhead: I replaced it with a simple related field

celm1990 (approved):
LGTM
@cormaza could you please review this?

PR #968, opened 2026-08-11
[19.0][FIX] account_payment_return_import_iso20022: Remove second search + code adaptation

pedrobaeza:
Follow-up of #883

It can produce false matches, as stated in #883.

@Tecnativa


pedrobaeza:
Equaling both branches:

/ocabot merge patch

PR #970, opened 2026-08-17
[18.0][IMP] account_payment_line: analytic distribution on counterpart lines

rjaraspearhead:
### Context
The payment counterpart line had a single **Analytic Account** field, but it
was **never propagated** to the journal item generated on posting, so the
analytic information was effectively lost.

### Changes
- The counterpart line now inherits `analytic.mixin`: the `analytic_account_id`
  field is replaced by an **`analytic_distribution`** (widget with % **and**
  amount, based on the line amount).
- On post, the distribution is written to the counterpart journal item(s) in
  `_prepare_move_line_counterpart_vals`.

### Migration
`migrations/18.0.1.1.0/post-migration.py` converts every existing line that had
an analytic account into a **100% distribution** to that account
(`{account_id: 100}`) and drops the legacy column. Tested end to end on a
throwaway database: a line with `analytic_account_id = 1` becomes
`analytic_distribution = {"1": 100}` and the old column is removed.

### Tests
Added `test_12_analytic_distribution`: the distribution set on the counterpart
line reaches the generated journal item.

@ChrisOForgeFlow, @celm1990, @OCA/accounting-maintainers


celm1990 (approved):
LGTM.

Just an existential question: the migration script converts the many2one field to JSON, but it does not reprocess the history in `account.move.line` to set the analytic_distribution and create the corresponding `account.analytic.line` records.

I'm not sure if we should reprocess the history to keep consistency between the records created before and after this commit.

@cormaza @xaviedoanhduy @Saran440 Could you please review this?

PR #972, opened 2026-08-18
[18.0][IMP] account_payment_notification: add automatic batch payment notifications

Ricardo-MC:
The goal of this change is to extend the existing account_payment_notification module to support automatic payment notifications based on company-specific configuration.

Currently, the module provides the notification mechanisms and allows payments to be notified manually or automatically when they are marked as sent. However, we need an automated process that can periodically identify payments that should be notified, according to each company's business rules.

The proposed change adds:

A configurable payment domain at company level, using the domain widget, so users can define additional criteria without modifying the code.
A payment notification start date, so users can define from which date payments should be considered without having to add a date condition manually to the domain.
A scheduled job that processes companies with automatic payment notifications enabled and uses the existing notification flow provided by this module.
Fixed domain conditions to ensure that only payments belonging to the company being processed and payments that are neither draft nor cancelled are considered.

We intentionally reuse the existing account_payment_notification_automatic, account_payment_notification_method, mark_as_sent(), and notification methods instead of introducing a parallel notification mechanism.

The intention is to keep the implementation generic: the module does not assume that notifications are only for vendors or for a specific payment type. The company-configured domain determines which payments should be notified.


account_payment_notification/data/ir_cron.xml
```diff
@@ -0,0 +1,12 @@
...
+<odoo noupdate="1">
+    <record id="ir_cron_notify_paid_batch_payments" model="ir.cron">
+        <field name="name">Notify Paid Batch Payments</field>
+        <field name="model_id" ref="account.model_account_payment" />
+        <field name="state">code</field>
+        <field name="code">model._cron_notify_paid_batch_payments()</field>
+        <field name="interval_number">1</field>
+        <field name="interval_type">hours</field>
+        <field name="active">True</field>
+    </record>
```
celm1990, line 11: Please set the nextcall to tomorrow morning.
```suggestion
    <record id="ir_cron_notify_paid_batch_payments" model="ir.cron">
        <field name="name">Notify Paid Payments</field>
        <field name="model_id" ref="account.model_account_payment" />
        <field name="state">code</field>
        <field name="code">model._cron_notify_paid_batch_payments()</field>
        <field name="interval_number">1</field>
        <field name="interval_type">days</field>
        <field name="active">True</field>
    </record>
```

account_payment_notification/models/account_payment.py
```diff
@@ -67,3 +76,44 @@ def _notify_sent_payments_sms(self):
...
             )
+
+    def _notify_payment_with_template(self, template):
+        """Notify a payment using the configured email template."""
+        self.ensure_one()
+        mt_comment = self.env.ref("mail.mt_comment")
+        if template.model != self._name:
+            raise ValidationError(
+                _("The selected email template must be for account.payment.")
+            )
```
celm1990: Is this case possible if you only allow selecting `mail templates` that belong to `account.payment`?

account_payment_notification/models/res_company.py
```diff
@@ -37,3 +38,30 @@ class ResCompany(models.Model):
             "notified using the chosen method(s)."
         ),
     )
+    account_payment_notification_template_id = fields.Many2one(
+        "mail.template",
+        string="Payment notification template",
+        domain=[("model", "=", "account.payment")],
+        help="Email template used to notify payments automatically.",
+    )
+    account_payment_notification_domain = fields.Text(
```
celm1990:
```suggestion
    account_payment_notification_domain = fields.Char(
```

account_payment_notification/models/res_company.py
```diff
@@ -37,3 +38,30 @@ class ResCompany(models.Model):
...
+        self.ensure_one()
+        base_domain = [
+            ("state", "=", "paid"),
+            ("batch_payment_id", "!=", False),
+            ("account_payment_notification_date", "=", False),
+        ]
+        if not self.account_payment_notification_domain:
+            return base_domain
+        configured_domain = safe_eval(self.account_payment_notification_domain)
+        return base_domain + list(configured_domain)
```
celm1990: The `batch_payment_id` field only exists in the Enterprise edition.

```suggestion
    def _get_account_payment_notification_domain(self):
        """Return the domain used to select payments for notification."""
        self.ensure_one()
        base_domain = [
            ("state", "=", "paid"),
            ("company_id", "=", self.id),
            ("account_payment_notification_date", "=", False),
        ]
        if not self.account_payment_notification_domain:
            return base_domain
        configured_domain = safe_eval(self.account_payment_notification_domain)
        return base_domain + list(configured_domain)
```

account_payment_notification/__manifest__.py
```diff
@@ -14,10 +14,12 @@
     "depends": [
         "account",
         "sms",
+        "account_payment_mode",
```
celm1990: Why is this dependency necessary?

celm1990 (changes requested):

account_payment_notification/wizards/res_config_settings_views.xml
```diff
@@ -29,6 +29,30 @@
...
+                    <field name="account_payment_notification_template_id" />
+                </setting>
+                <setting
+                    id="account_payment_notification_domain"
+                    company_dependent="1"
+                    help="Domain used to select payments for automatic notification."
+                >
+                    <field
+                        name="account_payment_notification_domain"
+                        placeholder="[('payment_type', '=', 'outbound')]"
```
celm1990, line 46:
```suggestion
                        name="account_payment_notification_domain"
                        placeholder="[('payment_type', '=', 'outbound')]"
                        widget="domain"
                        options="{'foldable': True, 'model': 'account.payment', 'in_dialog': True}"
```

celm1990 (changes requested):

account_payment_notification/wizards/res_config_settings_views.xml
Ricardo-MC: ping @celm1990

account_payment_notification/__manifest__.py
```diff
@@ -9,7 +9,7 @@
     "website": "https://github.com/OCA/account-payment",
     "author": "Moduon, Odoo Community Association (OCA)",
     "maintainers": ["yajo", "rafaelbn"],
-    "license": "LGPL-3",
+    "license": "AGPL-3",
```
celm1990, line 12: Please don't change the license.

account_payment_notification/models/res_company.py
```diff
@@ -37,3 +40,47 @@ class ResCompany(models.Model):
...
+                        limit=1,
+                    )
+            except Exception as error:
+                raise ValidationError(
+                    _(
+                        "Payment notification domain for company "
+                        "%(company)s is incorrectly formatted.",
+                        company=company.display_name,
+                    )
+                ) from error
```
celm1990, line 78: I think this is not necessary. With widget="domain", Odoo only allows users to select valid fields. If a user edits the domain manually, I think they must be a technical user. I removed this constraint to keep the code simple.

account_payment_notification/models/res_company.py
```diff
@@ -37,3 +40,47 @@ class ResCompany(models.Model):
...
+
+    def _get_account_payment_notification_domain(self):
+        """Return the configured domain for payment notification."""
+        self.ensure_one()
+        domain = literal_eval(self.account_payment_notification_domain or "[]")
+        return list(domain) + [
+            ("account_payment_notification_date", "=", False),
+        ]
```
celm1990, line 86: Please add a specific function to add the static domain so that it can be easily inherited. Additionally, the static domain should add these two conditions:

```
("state", "not in", ["draft", "cancel"])
("is_sent", "=", False)
("company_id", "=", self.id)
```

account_payment_notification/tests/test_notifications.py
```diff
@@ -240,3 +250,109 @@ def test_multilang(self):
...
+        )
+        self.assertEqual(len(messages), 1)
+        self.assertEqual(messages.subject, "Custom payment notification")
+
+    def test_invalid_notification_domain(self):
+        """Reject an invalid payment notification domain."""
+        with self.assertRaises(ValidationError):
+            self.company.sudo().account_payment_notification_domain = (
+                "[('invalid_field', '=', True)]"
+            )
```
celm1990, line 358: I see many tests with duplicated code. Please keep the minimum number of tests needed to cover the flow. By removing the constraints, some tests will no longer be necessary and can be removed as well.

Other tests can be integrated into existing tests.

celm1990 (changes requested):

account_payment_notification/models/account_payment.py
```diff
@@ -1,15 +1,24 @@
...
 class AccountPayment(models.Model):
     _inherit = "account.payment"
 
     # Used by sms.composer to get default phone
     mobile = fields.Char(related="partner_id.mobile")
+    account_payment_notification_date = fields.Datetime(
+        string="Payment Notification Date",
+        copy=False,
+        readonly=True,
+    )
```
celm1990, line 21: Review the Odoo core and the account_payment_notification module. I think there is no need for a new field to differentiate whether the payment has been sent or not. The payment already has the `is_sent` field, so please reuse it.

https://github.com/odoo/odoo/blob/a47a80fb04d113e3da83e654b667e7335cf63d39/addons/account/models/account_payment.py#L1158-L1162

account_payment_notification/models/res_company.py
```diff
@@ -37,3 +40,47 @@ class ResCompany(models.Model):
             "notified using the chosen method(s)."
         ),
     )
+    account_payment_notification_template_id = fields.Many2one(
+        "mail.template",
+        string="Payment notification template",
+        domain=[("model", "=", "account.payment")],
+        help="Email template used to notify payments automatically.",
+    )
```
celm1990, line 48: Reuse the existing fields in this module. There is no need for a field to select the mail template. This module already creates a mail template, so use it by its XML ID.
https://github.com/OCA/account-payment/blob/bb0836e1cc02e2c40cf70fac10d1c53a7f2c28d2/account_payment_notification/models/account_payment.py#L57

Image


```suggestion
```

account_payment_notification/models/res_company.py
```diff
@@ -37,3 +40,47 @@ class ResCompany(models.Model):
...
+    account_payment_notification_domain = fields.Char(
+        string="Payment notification domain",
+        default="[('state', 'not in', ['draft', 'cancel'])]",
+        help="Domain used to select payments for automatic notification.",
+    )
+    account_payment_notification_cron = fields.Boolean(
+        string="Automatic payment notifications",
+        help="Enable automatic payment notifications.",
+    )
```
celm1990, line 57: The same applies here. This module already has a field to determine whether automatic notifications are enabled: `account_payment_notification_automatic`.
https://github.com/OCA/account-payment/blob/bb0836e1cc02e2c40cf70fac10d1c53a7f2c28d2/account_payment_notification/models/res_company.py#L10


```suggestion
```

account_payment_notification/models/account_payment.py
```diff
@@ -67,3 +76,40 @@ def _notify_sent_payments_sms(self):
...
+                        company.account_payment_notification_template_id
+                    )
+                    payment.account_payment_notification_date = fields.Datetime.now()
+            except Exception:
+                _logger.exception(
+                    "Error while processing payment notifications "
+                    "for company %s (%s).",
+                    company.display_name,
+                    company.id,
+                )
```
celm1990, line 115: In short, if we reuse the fields and logic that already exist in this module, the cron only needs to search for the payments and call the `mark_as_sent` function. The module already implements the required logic, so nothing else is needed. This approach significantly reduces the amount of code.

 https://github.com/OCA/account-payment/blob/bb0836e1cc02e2c40cf70fac10d1c53a7f2c28d2/account_payment_notification/models/account_payment.py#L14


```suggestion
    @api.model
    def _cron_notify_paid_batch_payments(self):
        """Notify paid payments belonging to payment batches."""
        companies = self.env["res.company"].search(
            [
                ("account_payment_notification_automatic", "=", "auto")
            ]
        )
        for company in companies:
                domain = company._get_account_payment_notification_domain()
                payments = (
                    self.env["account.payment"].with_company(company).search(domain)
                )
                for payment in payments:
                    payment.mark_as_sent()
```

celm1990 (changes requested):
Please review my comments. This change should be simple; we are making it unnecessarily complex. Let's reuse the fields and logic that this module already has.

If you have any doubts, please let me know.

Ricardo-MC:
ping @celm1990 

> Please review my comments. This change should be simple; we are making it unnecessarily complex. Let's reuse the fields and logic that this module already has.
> 
> If you have any doubts, please let me know.

account_payment_notification/models/res_company.py
```diff
@@ -37,3 +39,24 @@ class ResCompany(models.Model):
             "notified using the chosen method(s)."
         ),
     )
+    account_payment_notification_domain = fields.Char(
+        string="Payment notification domain",
+        default="[('state', 'not in', ['draft', 'cancel'])]",
```
celm1990: This default is redundant with the static domain set in the `_get_account_payment_notification_static_domain` method. Please remove it from here and keep it only in `_get_account_payment_notification_static_domain`.

```suggestion
        default="[]",
```

celm1990 (changes requested):
Just a minor change

account_payment_notification/models/res_company.py
Ricardo-MC: @celm1990

yajo (changes requested):
Would you please elaborate a description that explains what problem are you solving and why are you proposing this change?

Including STR, screenshots or screencasts about the problem helps a lot to understand it.

Ricardo-MC:
> Would you please elaborate a description that explains what problem are you solving and why are you proposing this change?
> 
> Including STR, screenshots or screencasts about the problem helps a lot to understand it.

Thanks for the review

rafaelbn:
Hello @Ricardo-MC ,

Thank you for your contribution.

IMHO you should create a new module which depends on `account_payment_notification` for this extended use case. Why? Because this make a product much more complicated to use for everyone using this module and this is a not desired objetive.

`account_payment_notification_extended` (could be an option with a separate an independent README, USAGE, CONFIGURATION, etc.

Best regards,
Rafael

PR #974, opened 2026-08-30
[19.0][MIG] account_payment_multi_deduction: Migration to 19.0

amincheloh:
Migration of `account_payment_multi_deduction` to 19.0.

This module is not listed in #885 — it was migrated to 18.0 in April 2026, after that issue was generated, so the bot never picked it up. Happy for a maintainer to add it to the list.

### Changes
- Bumped version to `19.0.1.0.0`
- Removed the `domain=[("deprecated", "=", False)]` on `account.payment.deduction.account_id`. `account.account.deprecated` no longer exists in 19.0 — it was replaced by the standard `active` field (17.0 has `deprecated` and no `active`; 19.0 has `active` and no `deprecated`). Without this the module fails to install:
  ```
  Unknown field "account.account.deprecated" in domain of python field 'account_id'
  ```
  No replacement domain is needed: `active` is filtered automatically, and core's own `writeoff_account_id` carries no domain either.
- Replaced an f-string inside `self.env._()` with a named `%(...)s` parameter. It produced no `.pot` entry and is flagged by `translation-fstring-interpolation` in `.pylintrc-mandatory`

### Testing
Installed and tested on 19.0.0. All 4 existing tests pass with demo data enabled (matching `oca_install_addons`), and the three view xpaths into core's payment register form still resolve. Module coverage is 99% (`account_payment_register.py` and `account_payment_deduction.py` both at 100%).


amincheloh:
Hello @Saran440, could you please review this migration

